PR #71, opened 2017-12-09
The "minor changes" PR

Ampli-fier:
The only **fix** is:  swap arguments for the _addcontact_ command in cmdline.
The rest is just cosmetics :)

### DIFF
--- a/cmdline/cmdline.c
+++ b/cmdline/cmdline.c
@@ -310,7 +310,7 @@ char* mrmailbox_cmdline(mrmailbox_t* mailbox, const char* cmdline)
 		if( arg1 && strcmp(arg1, "imex")==0 )
 		{
 			ret = safe_strdup(
-				"====================Import-/export-commands==\n"
+				"====================Import/Export commands==\n"
 				"initiate-key-transfer\n"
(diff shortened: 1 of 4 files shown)


r10s:
You're becoming a C-Expert :)

Ampli-fier:
Ha ha :)
Actually, my current attempts to compile C-code do no end any more in **segmentation faults** ... sometimes ...

r10s:
:)

Ampli-fier:
I noticed an error in my commit for swapping arguments for addcontact.

In fact, the first parameter, which is the name is optional and must be mentioned first but with brackets [ ].
So this is the correct syntax for the two lines in cmdline.c:

`"addcontact [<name>] <addr>\n"`

`ret = safe_strdup("ERROR: Arguments [<name>] <addr> expected.");`

Can you please update the code accordingly?

r10s:
> Can you please update the code accordingly?

done so, see https://github.com/deltachat/deltachat-core/commit/f14c518b70995dcbdf49f86db5208d93f1b1cb6c (being around there anyway :)

PR #82, opened 2017-12-29
Update build instructions for meson

flub:
Also clean up the build file somewhat.  No need to leave cruft around.


r10s:
perfect, just wanted to ask you some things, now trying over :)

PR #83, opened 2017-12-30
WIP: First step towards python bindings

flub:
DO NOT MERGE

This is a very rough first stab at compiling CFFI Python bindings for deltachat-core.  It compiles and generates some bindings, but that's about it.

1. I've done this as a PR in this repo instead of a new one.  It makes header modifications easier and means I don't have to worry about copyrights.  It may even make sense to leave the raw bindings here but it can always be moved out later.

2. This runs the header files through the preprocessor (`gcc -E`) using the `PY_CFFI` symbol to end up with something which is parsable by cffi.  I think this will stay but the implementation cleaned up somewhat and the number of ifdefs reduced.

3. I've effectively made the structs opaque/private for cffi.  I did this as otherwise I had to worry about defining opaque structs for pthread_t and related friends.  Since you consider them private anyway maybe it's better to do that right away and properly instead of my cffi-specific hack here (I guess with `typedef _mrmailbox mrmailbox_t;` + mrmailbox-private.h etc?)

4. I've chopped up the included headers with ifdef.  That's just temporarily, once we have opaque structs and all public.h files are parsable by cffi only the external includes need to be ifdef'ed out.

5. I have no idea what `clist` is supposed to be and where it comes from.  I've defined it as an opaque struct and it seems to work.  I'd know better what to do with it if I knew more about it.

6. Mostly as an aside, some header files seem to have incomplete includes, e.g. `mrmsg.h` uses `time_t` but doesn't include `time.h`.

### DIFF
--- a/python/src/deltachat/_build.py
+++ b/python/src/deltachat/_build.py
@@ -1,23 +1,35 @@
-import subprocess
+import distutils.ccompiler
+import distutils.sysconfig
 import tempfile
 
 import cffi
 
 
-ffibuilder = cffi.FFI()
-ffibuilder.set_source(
-    'deltachat.capi',
-    """
-    #include <deltachat/mrmailbox.h>
-    """,
-    libraries=['deltachat'],
-)
-with tempfile.NamedTemporaryFile(mode='r') as fp:
-    proc = subprocess.run(['gcc', '-E', '-o', fp.name, '-DPY_CFFI=1',
-                           '../src/mrmailbox.h'])
-    proc.check_returncode()
-    ffibuilder.cdef(fp.read())
+def ffibuilder():
+    builder = cffi.FFI()
+    builder.set_source(
+        'deltachat.capi',
+        """
+            #include <deltachat/mrmailbox.h>
+        """,
+        libraries=['deltachat'],
+    )
+    builder.cdef("""
+        typedef int... time_t;
+    """)
+    cc = distutils.ccompiler.new_compiler(force=True)
+    distutils.sysconfig.customize_compiler(cc)
+    with tempfile.NamedTemporaryFile(mode='w', suffix='.h') as src_fp:
+        src_fp.write('#include <deltachat/mrmailbox.h>')
+        src_fp.flush()
+        with tempfile.NamedTemporaryFile(mode='r') as dst_fp:
+            cc.preprocess(source=src_fp.name,
+                          output_file=dst_fp.name,
+                          macros=[('PY_CFFI', '1')])
+            builder.cdef(dst_fp.read())
+    return builder
 
 
 if __name__ == '__main__':
-    ffibuilder.compile(verbose=True)
+    builder = ffibuilder()
+    builder.compile(verbose=True)
--- a/python/src/deltachat/_build.py
+++ b/python/src/deltachat/_build.py
@@ -1,0 +1,23 @@
+import subprocess
+import tempfile
+
+import cffi
+
+
+ffibuilder = cffi.FFI()
+ffibuilder.set_source(
+    'deltachat.capi',
+    """
+    #include <deltachat/mrmailbox.h>
+    """,
+    libraries=['deltachat'],
+)
+with tempfile.NamedTemporaryFile(mode='r') as fp:
+    proc = subprocess.run(['gcc', '-E', '-o', fp.name, '-DPY_CFFI=1',
+                           '../src/mrmailbox.h'])
+    proc.check_returncode()
+    ffibuilder.cdef(fp.read())
+
+
+if __name__ == '__main__':
+    ffibuilder.compile(verbose=True)
--- a/python/src/deltachat/_build.py
+++ b/python/src/deltachat/_build.py
@@ -16,6 +16,7 @@ def ffibuilder():
     )
     builder.cdef("""
         typedef int... time_t;
+        void free(void *ptr);
     """)
     cc = distutils.ccompiler.new_compiler(force=True)
     distutils.sysconfig.customize_compiler(cc)
--- a/src/mrmailbox.h
+++ b/src/mrmailbox.h
@@ -150,18 +150,15 @@ extern "C" {
 #ifndef PY_CFFI
 #include <pthread.h>
 #include <libetpan/libetpan.h> /* defines uint16_t */
+#endif
+
 #include "mrarray.h"
-#endif
 #include "mrchatlist.h"
-#ifndef PY_CFFI
 #include "mrchat.h"
-#endif
 #include "mrmsg.h"
 #include "mrcontact.h"
-#ifndef PY_CFFI
 #include "mrlot.h"
 #include "mrevent.h"
-#endif
 
 
 /**
--- a/src/mrmailbox.h
+++ b/src/mrmailbox.h
@@ -147,15 +147,21 @@ extern "C" {
  */
 
 
+#ifndef PY_CFFI
 #include <pthread.h>
 #include <libetpan/libetpan.h> /* defines uint16_t */
 #include "mrarray.h"
+#endif
 #include "mrchatlist.h"
+#ifndef PY_CFFI
 #include "mrchat.h"
+#endif
 #include "mrmsg.h"
 #include "mrcontact.h"
+#ifndef PY_CFFI
 #include "mrlot.h"
 #include "mrevent.h"
+#endif
 
 
 /**
--- a/src/mrmsg.h
+++ b/src/mrmsg.h
@@ -26,6 +26,10 @@
 extern "C" {
 #endif
 
+#ifdef PY_CFFI
+typedef int... time_t;
+#endif
+
 
 typedef struct _mrmailbox mrmailbox_t;
 
--- a/src/mrmsg.h
+++ b/src/mrmsg.h
@@ -26,10 +26,6 @@
 extern "C" {
 #endif
 
-#ifdef PY_CFFI
-typedef int... time_t;
-#endif
-
 
 typedef struct _mrmailbox mrmailbox_t;
 
--- a/python/setup.py
+++ b/python/setup.py
@@ -1,0 +1,22 @@
+import setuptools
+
+
+setuptools.setup(
+    name='deltachat',
+    version='0.1',
+    description='Python bindings for deltachat-core using CFFI',
+    author='Delta Chat contributors',
+    setup_requires=['cffi>=1.0.0'],
+    install_requires=['cffi>=1.0.0'],
+    packages=setuptools.find_packages('src'),
+    package_dir={'': 'src'},
+    cffi_modules=['src/deltachat/_build.py:ffibuilder'],
+    classifiers=[
+        'Development Status :: 3 - Alpha',
+        'Intended Audience :: Developers',
+        'License :: OSI Approved :: GNU General Public License (GPL)',
+        'Programming Language :: Python :: 3',
+        'Topic :: Communications :: Email',
+        'Topic :: Software Development :: Libraries',
+    ],
+)


python/src/deltachat/_build.py
```diff
@@ -0,0 +1,23 @@
...
+ffibuilder.set_source(
+    'deltachat.capi',
+    """
+    #include <deltachat/mrmailbox.h>
+    """,
+    libraries=['deltachat'],
+)
+with tempfile.NamedTemporaryFile(mode='r') as fp:
+    proc = subprocess.run(['gcc', '-E', '-o', fp.name, '-DPY_CFFI=1',
+                           '../src/mrmailbox.h'])
```
flub: Clean up this implementation, e.g. don't hardcode gcc, don't hardcode path to header file.

src/mrmsg.h
```diff
@@ -26,6 +26,10 @@
 extern "C" {
 #endif
 
+#ifdef PY_CFFI
+typedef int... time_t;
+#endif
```
flub: Move external declarations to `_build.py`, no need to pollute the actual headers with this CFFI detail.

src/mrmailbox.h
```diff
@@ -346,7 +355,11 @@ void            mrmailbox_log_error         (mrmailbox_t*, int code, const char*
 void            mrmailbox_log_error_if      (int* condition, mrmailbox_t*, int code, const char* msg, ...);
 void            mrmailbox_log_warning       (mrmailbox_t*, int code, const char* msg, ...);
 void            mrmailbox_log_info          (mrmailbox_t*, int code, const char* msg, ...);
+#ifndef PY_CFFI
 void            mrmailbox_log_vprintf       (mrmailbox_t*, int event, int code, const char* msg, va_list);
+#else
+void            mrmailbox_log_vprintf       (mrmailbox_t*, int event, int code, const char* msg, ...);
+#endif
```
flub: Figure out what's up with va_list.  AFAIK the `...` notation is also valid C but it's what CFFI can parse.  This results in a compilation warning though:
```
warning: assignment from incompatible pointer type [-Wincompatible-pointer-types]
   *(void(* *)(mrmailbox_t *, int, int, char const *, ...))o = mrmailbox_log_vprintf;
```

flub:
So as next step I'd think that making the structs opaque/private is the best thing to do and once that's done I could come back and update/improve this.  What do you think?

src/mrmailbox.h
r10s: maybe the non-`...` notation also works for the other systems; we can then switch completely.

r10s:
> 1. I've done this as a PR in this repo instead of a new one.  It makes header modifications easier and means I don't have to worry about copyrights.  It may even make sense to leave the raw bindings here but it can always be moved out later.

Okay, I also think this is easier for now. If we encounter any disadvantages, we can move it outside.

> 2. This runs the header files through the preprocessor (`gcc -E`) using the `PY_CFFI` symbol to end up with something which is parsable by cffi.  I think this will stay but the implementation cleaned up somewhat and the number of ifdefs reduced.

I have not tried this out yet, never used cffi, but i think a #ifdef here and there is okay.

> 3. I've effectively made the structs opaque/private for cffi.  I did this as otherwise I had to worry about defining opaque structs for pthread_t and related friends.  Since you consider them private anyway maybe it's better to do that right away and properly instead of my cffi-specific hack here (I guess with `typedef _mrmailbox mrmailbox_t;` + mrmailbox-private.h etc?)

Yes, it may be better if the core itself declares the structs in a opaque/private way. This may also make other bindings easier. I have tried it out with mrchat_t. I am wondering if it will be sufficient for CFFI to have only a forward declaration as 
```
typedef struct mrchat_t mrchat_t;
```
(using the same name in the struct and typedef namespaces) Internally I can include `mrchat-internal.h` then. Works for me.

However, If you see any disadvantages here, we can also use `typedef _mrchat mrchat_t`, maybe this also makes things clearer.

In any case an explicit "public" declaration as `typedef struct mrchat_t  { ... } mrchat_t` should not be needed then.

See the branch https://github.com/deltachat/deltachat-core/tree/mrchat-private for details.

If we have figured out what works best for all (known) cases, I will declare the other sructures the same way. You can then continue with CFFI.

> 4. I've chopped up the included headers with ifdef.  That's just temporarily, once we have opaque structs and all public.h files are parsable by cffi only the external includes need to be ifdef'ed out.

Okay.

> 5. I have no idea what `clist` is supposed to be and where it comes from.  I've defined it as an opaque struct and it seems to work.  I'd know better what to do with it if I knew more about it.

clist is only used internally, so it is fine to declare it as "anything".

> 6. Mostly as an aside, some header files seem to have incomplete includes, e.g. `mrmsg.h` uses `time_t` but doesn't include `time.h`.

The header hell ... feel free to add headers as you encounter they're missing for CFFI.

Cheers - und einen "guten Rutsch" nachher :)

src/mrchatlist.h
```diff
@@ -42,12 +42,16 @@ typedef struct mrarray_t    mrarray_t;
  */
 typedef struct mrchatlist_t
 {
+#ifndef PY_CFFI
```
r10s: maybe we can git rid of these #ifdef's see my other comment.

src/mrmsg.h
r10s: well, if time.h is missing for CFFI it may be also missing for other systems. So, i think it is fine just to include <time.h>. But you are right, it's only a detail.

flub: I think you misunderstand.  The code does and should include time.h.  However this header file is a bit complicated for the cffi parser so I'd guard the import -together with all other external headers- with an #ifdef PY_CFFI and thus not include it for cffi.  However cffi still needs to know that time_t is some kind of integer, which is what this "declaration" does (cffi then uses the compiler to work out exactly what size this integer is etc).  Hence this cffi-specific declaration of time_t should be moved to _build.py.

r10s: okay, I see :) - and a happy new year :)

karissa:
Can you move it outside of the repo? I am also making Node bindings and I think it's quite bulky to have to download the python code in order to use it. What I'm doing in my repo is making deltachat-core a git submodule of my node bindings, so you'll just have to fix your paths in your python code and voila!

r10s:
@karissa thank you for your post, I also think that the clear solution is a separate repository; however, for now, it seems to be easier to leave it here and move it away as soon as appropriate :)

src/mrmailbox.h
flub: I'm not sure I understand your response here.  I think the standard C-notation for a function with variadic arguments is `f(int named, ...)`.  If I understand correctly you pass in va_list as a lone type declaration in the signature which afaik is not standard C.  So is there a reason to do this instead of the standard `...` notationn?  It would be nice if all code could use the standard notation for variadic function arguments instead of having to ifdef for cffi like this.  What would you lose if you moved all varargs to this?

r10s: > So is there a reason to do this instead of the standard ... notationn? 

Yes, the `...`-notation does not allow to pass a variable argument list to another function. The solution for this is to convert the list into an object using va_start() and pass this to other functions. 

However, there seems to be an easy solution for us: I think I can make mrmailbox_log_vprintf() private so that it is not seen by CFFI.

NB: I am currently on the way moving all structures to internal files as `struct _mrchat { ... }` the public headers use a `typedef struct _mrchat mrchat_t;` then.

flub:
Björn Petersen <notifications@github.com> writes:
>> 3. I've effectively made the structs opaque/private for cffi.  I did this as otherwise I had to worry about defining opaque structs for pthread_t and related friends.  Since you consider them private anyway maybe it's better to do that right away and properly instead of my cffi-specific hack here (I guess with `typedef _mrmailbox mrmailbox_t;` + mrmailbox-private.h etc?)
>
> Yes, it may be better if the core itself declares the structs in a opaque/private way. This may also make other bindings easier. I have tried it out with mrchat_t. I am wondering if it will be sufficient for CFFI to have only a forward declaration as 
> ```
> typedef struct mrchat_t mrchat_t;
> ```
> (using the same name in the struct and typedef namespaces) Internally I can include `mrchat-internal.h` then. Works for me.
>
> However, If you see any disadvantages here, we can also use `typedef _mrchat mrchat_t`, maybe this also makes things clearer.
>
> In any case an explicit "public" declaration as `typedef struct mrchat_t  { ... } mrchat_t` should not be needed then.
>
> See the branch https://github.com/deltachat/deltachat-core/tree/mrchat-private for details.

Ok, I've tried this branch and the `typedef struct mrchat_ mrchat_t;`
works fine for CFFI.  Would be great if you could update master with
this everywhere.


> Cheers - und einen "guten Rutsch" nachher :)

Happy new year to you too!

flub:
Karissa McKelvey <notifications@github.com> writes:

> Can you move it outside of the repo? I am also making Node bindings
> and I think it's quite bulky to have to download the python code in
> order to use it. What I'm doing in my repo is making deltachat-core a
> git submodule of my node bindings, so you'll just have to fix your
> paths in your python code and voila!

A submodule seems like it could be nice too.  Maybe that's something to
move to.

I don't think this will ever be a bulky download though.  It's not much
code and can be completely ignored when just building the
libdeltachat.so.

I'm somewhat in favour of doing the PR against deltachat-core before
deciding to split it off because I'm legally not allowed to just create
a new repo by my employer.  There's several other workarounds this too
probably.

I'll continue to evolve this PR but will check back before asking to
merge to see if we'd like to merge here or move to a separate repo right
away.

r10s:
> Ok, I've tried this branch and the `typedef struct mrchat_ mrchat_t;` works fine for CFFI. Would be great if you could update master with this everywhere.

I'm just doing it now, however, it may take a moment :)

r10s:
@flub okay, the main headers (mrmailbox.h and headers included from there) use private opaque/structure now. As a side effect, less dependencies are needed, eg. types as mrimap_t or mrsqlite_t are no longer needed. the reference to va_list is also removed. 

@karissa do the changes make any problems to the JavaScript binding?

karissa:
@r10s javascript binding is still a wip, it's quite flexible and can adapt to these changes. I will let you know if I find something else that needs to be changed.

flub:
This has now been updated and compiles the bindings with only a minor change to one header file.

It also now uses the installed header version, given it also compiles against the installed version this might make sense.  Build instructions are something like:
```
meson builddir
cd builddir
ninja
ninja install
cd python
pip install [-e] .
python -c 'import deltachat.capi; print(dir(deltachat.capi.lib))'
```

So should this build against an installed version?  If so there's little point in having this in the deltachat-core repo I guess.  However if these bindings should also expose internal APIs for testing purposes maybe it makes sense to have them here anyway, and maybe it makes sense to figure out how to build against the non-installed version.  Thoughts?

/cc @hpk42

hpk42:
On Thu, Jan 04, 2018 at 14:22 -0800, Floris Bruynooghe wrote:
> This has now been updated and compiles the bindings with only a minor change to one header file.
> 
> It also now uses the installed header version, given it also compiles against the installed version this might make sense.  Build instructions are something like:
> ```
> meson builddir
> cd builddir
> ninja
> ninja install
> cd python
> pip install [-e] .
> python -c 'import deltachat.capi; print(dir(deltachat.capi.lib))'
> ```
> 
> So should this build against an installed version?  If so there's little point in having this in the deltachat-core repo I guess.  However if these bindings should also expose internal APIs for testing purposes maybe it makes sense to have them here anyway, and maybe it makes sense to figure out how to build against the non-installed version.  Thoughts?
> 
> /cc @hpk42 

not sure i fully understand the question but ...

i guess the first immediately useful thing would be to make the above into a travis script
and pip-install pytest and then run tests (a first one could be accessing deltachat.capi.lib). 

Creating an installable python package which includes the deltachat c and ffi code is a somewhat separate effort, isn't it?

r10s:
I think this can be merged now? Beside the two lines in mrmailbox.h there are no further changes required.

And regarding these two lines: `pthread.h` is no longer required here and the `libetpan.h` can be replaced by `inttypes.h`+`time.h` - while time_t seems to make problems, uint32_t etc. seems just to work in python? Maybe we should replace the few occurrences of time_t by int64_t to solve this issue and have fewer dependencies in the header.

flub:
I'm not really convinced this is what you would like to get out of this.  If I understood correctly the purpose of doing this is so that you can write tests against libdeltachat in python.  This PR currently allows you to build a separate binding, from an *installed* libdeltachat.  This is nice if the goal is to have Python bindings for other users, but rather cumbersome for a TDD loop.  (and sure TDD is an extreme here, but it brings the point across.  Same for fixing any bug identified by a test)

@r10s Wrt imports, would you like to get rid of the #ifndef PY_CFFI bits in your headers all together?  For me the current situation seems very sane.  I don't think in the long run getting rid of this is likely anyway as you want to test internal APIs, so might end up pulling the private headers into CFFI anyway, at which point they'll need this as well.

I'm not a particular fan of obscuring time_t by int64, the current way time_t is handled in these bindings is perfectly fine to me.

r10s:
@flub I am fine with the `#ifndef PY_CFFI`. From my point of view, this can be merged if I get the okay from you - still discourage by the DO NOT MERGE atop ...

To the python-related questions: I don't know - but probably @hpk will :)

hpk42:
On Sat, Jan 06, 2018 at 10:36 +0000, Floris Bruynooghe wrote:
> I'm not really convinced this is what you would like to get out of this.  If I understood correctly the purpose of doing this is so that you can write tests against libdeltachat in python.  This PR currently allows you to build a separate binding, from an *installed* libdeltachat.  This is nice if the goal is to have Python bindings for other users, but rather cumbersome for a TDD loop.  (and sure TDD is an extreme here, but it brings the point across.  Same for fixing any bug identified by a test)


the functional test suite i have in mind would be written
such that one builds and debugs test code without changing the C code
or the bindings.  All this test-writing and refining activity would
happen after the bindings are created.  Once a test fails that should
pass, we are entering test-driven-development (TDD) -- at which point
it's interesting to have fast compile -> run tests cycles.  Does the
language binding creation really add much overhead compared to the
overall involved build steps?

flub:
@r10s I'm fine with merging as-is now, It gives a bare-bones Python bindings.  Mind you the user will still be responsible for respecting the C semantics like figuring out the right calling conventions, calling `deltachat.capi.lib.free()` when required, using `deltachat.capi.ffi.string()` when needed etc etc.

@hpk42 So current bindings building and running against an installed libdeltachat are fine.  The step to work on in-development version is to make mason build the python extension as then it can tell us where the header files are and where the lib to link with is.  There's a python3 module in meson for this and some examples of pure-python, extension-modules and cython.  So figuring out how to do cffi from there should be possible.  I think this is the more correct thing to do here.  As it's not the most important there's no need to stop this from being used.

r10s:
Yeah! We have basic Python support now :) \o/

hpk42:
On Sat, Jan 06, 2018 at 21:42 +0000, Floris Bruynooghe wrote:
> @r10s I'm fine with merging as-is now, It gives a bare-bones Python bindings.  Mind you the user will still be responsible for respecting the C semantics like figuring out the right calling conventions, calling `deltachat.capi.lib.free()` when required, using `deltachat.capi.ffi.string()` when needed etc etc.
> 
> @hpk42 So current bindings building and running against an installed libdeltachat are fine.  The step to work on in-development version is to make mason build the python extension as then it can tell us where the header files are and where the lib to link with is.  There's a python3 module in meson for this and some examples of pure-python, extension-modules and cython.  So figuring out how to do cffi from there should be possible.  I think this is the more correct thing to do here.  As it's not the most important there's no need to stop this from being used.

thanks a lot for your work on this, @flub!  
maybe we find a weekend sometime to hack a bit on things and also
take walks together? 

FYI, right now, i am focusing my own hacking activities on
https://github.com/hpk42/py-autocrypt (soon to be renamed) -- lots of
things planned there, not the least mutt and mailing list integration
support, and a push towards Kappa architecture
(http://milinda.pathirage.org/kappa-architecture.com/) to make the flow
of cryptographic key material debuggable between peers.  It's something
i'd like to suggest to @r10s wrt Delta as well but first i need to
refactor ...

PR #87, opened 2018-01-05
Grammar changes

Ampli-fier:
Did not know the meaning, thus how to correct the following:
https://github.com/deltachat/deltachat-core/blob/f14c518b70995dcbdf49f86db5208d93f1b1cb6c/src/mrevent.h#L270

### DIFF
--- a/src/mrchat.c
+++ b/src/mrchat.c
@@ -119,7 +119,7 @@ void mrchat_empty(mrchat_t* chat)
  * Get chat ID. The chat ID is the ID under which the chat is filed in the database.
  *
  * Special IDs:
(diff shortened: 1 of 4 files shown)


r10s:
send sent sent
send sent sent
send sent sent
send sent sent
(I would write it down 100 more times, however, not sure if one learns grammar by using ctrl+v)

r10s:
> Before a critical function _MR_EVENT_WAKE_LOCK with data1=1_ is called, it it finishes, _MR_EVENT_WAKE_LOCK with data1=0_ is called. 

The meaning is that critical stuff is surrounded by MR_EVENT_WAKE_LOCK as follows:
```
MR_EVENT_WAKE_LOCK(data1=1):
crtical_stuff_that_must_not_be_aborted();
MR_EVENT_WAKE_LOCK(data1=0):
```

PR #108, opened 2018-02-22
Check msgid before delete

r10s:
- avoid deleting the wrong message if UIDVALIDITY has changed
- in the future, we may SEARCH for the message-id if folder/UID does not point to the correct one


r10s:
thanks for merging, follow up is #109

PR #109, opened 2018-02-23
if the UID was not found on deletion of a message, try to search the …

r10s:
…UID by the Message-ID (the message may eg. be moved around on another MUA)

### DIFF
--- a/src/mrimap.c
+++ b/src/mrimap.c
@@ -419,6 +419,45 @@ static void forget_folder_selection__(mrimap_t* ths)
 }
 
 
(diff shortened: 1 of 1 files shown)


r10s:
thanks for merging.

> Is the comment: "On failure, 0 is returned and any or none folder is selected." in search_uid__() correct? free_folders is called in cleanup.

this is okay, free_folders() just deletes the folder list in memory and does nothing on the IMAP server, this is only done when select_folder__() is called - and on failure it may happen that select_folder__() was never called or failed so that no folder is left selected on the IMAP server.

PR #133, opened 2018-03-12
patch makes meson create pkg-config file so apps can make use of it

mogorman:
I am working on a plugin using deltachat-core. this patch makes meson create a pkg-config file so my library can make use of it easily.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -40,3 +40,18 @@ subdir('src')
 
 # Build the binaries.
 subdir('cmdline')
+
+version = run_command('git', 'describe', '--tags')
+if version.returncode() != 0
+  version = 'unknown-error'
+else
+  version = version.stdout().strip()
+endif
+
+pkg = import('pkgconfig')
+pkg.generate(libraries : lib,
+             version : version,
+             subdirs : ['deltachat'],
+             name : 'libdeltachat',
+             filebase : 'deltachat',
+             description : ' Create your own, email-compatible messenger.')


mogorman:
This is still wip I need to have it properly announce it's dependencies

r10s:
@mogorman thank you for your patch, however, i have no idea what the changes do :) 
but i think sooner or later @hpk42 or @flub will have a look at it.

meson.build
```diff
@@ -40,3 +40,18 @@ subdir('src')
 
 # Build the binaries.
 subdir('cmdline')
+
+version = run_command('git', 'describe', '--tags')
+if version.returncode() != 0
+  version = 'unknown-error'
```
flub: If you're building without git maybe it would be better not to build the pkg-config file at all?  Otherwise this might get somewhat confusing.  OTOH debugging why you can't build the pkg-config file might also be annoying, maybe going `message('git version not found, pkg-config will not be generated')` is the right thing to do then?  What do you think?

flub:
Is there a reason you need pkg-config?  I guess if you're starting from scratch your app can also use meson and pull it in like that?

Not that I'm against a pkg-config file at all.  It's a great addition in any case!

mogorman:
I am working on making a libpurple plugin using the core as a library. It makes sense for the library to be discover able so my library can reference it as a dependency in the same way deltachat-core calls out to its dependencies...

flub:
ah, cool.  makes perfect sense.

mogorman:
I need to get the build completely working on my end. I think what I have is enough but will confirm this weekend.

PR #137, opened 2018-03-28
improve autoconfig

r10s:
search for autoconfig also in example.com/.well-known/autoconfig/mail dir

refers to  https://github.com/deltachat/deltachat-core/commit/5cee73a750091d3a20b645fd37ca7863573f3e3e#commitcomment-28308841

regarding the `emailaddress=` parameter:
the [doc](https://developer.mozilla.org/en-US/docs/Mozilla/Thunderbird/Autoconfiguration) does not mention it, however, Thunderbird uses it, at least in the [source code](https://releases.mozilla.org/pub/thunderbird/) 45.5.1. as it makes some sense, i do not think, this has changed: the .well-known dir is for users who cannot add a subdomain - why not give them the same options (to return eg. customized login names) as for the subdomain solution.

looking at the thunderbird source code, i am also wondering, why they do not look at the https domains, however, i have not dived deeply in the source code, maybe their http-request object just uses both variants or they rely on a redirection. i think, for delta, it makes some sense to use both variants.

### DIFF
--- a/src/mrmailbox.h
+++ b/src/mrmailbox.h
@@ -302,7 +302,7 @@ void            mrmailbox_heartbeat         (mrmailbox_t*);
 
 
 /* out-of-band verification */
-#define         MR_QR_FPR_ASK_OOB           200 /* id=contact, test2=return_tag */
+#define         MR_QR_FPR_ASK_OOB           200 /* id=contact, text2=random secret */
 #define         MR_QR_FPR_OK                210 /* id=contact */
 #define         MR_QR_FPR_MISMATCH          220 /* id=contact */
 #define         MR_QR_FPR_WITHOUT_ADDR      230 /* test1=formatted fingerprint */
--- a/src/mrmailbox.h
+++ b/src/mrmailbox.h
@@ -302,7 +302,7 @@ void            mrmailbox_heartbeat         (mrmailbox_t*);
 
 
 /* out-of-band verification */
-#define         MR_QR_FPR_ASK_OOB           200 /* id=contact, text2=random secret */
+#define         MR_QR_FPR_ASK_OOB           200 /* id=contact, text1=fingerprint, text2=random secret */
 #define         MR_QR_FPR_OK                210 /* id=contact */
 #define         MR_QR_FPR_MISMATCH          220 /* id=contact */
 #define         MR_QR_FPR_WITHOUT_ADDR      230 /* test1=formatted fingerprint */
--- a/src/mrmailbox_configure.c
+++ b/src/mrmailbox_configure.c
@@ -500,7 +500,16 @@ int mrmailbox_configure_and_connect(mrmailbox_t* mailbox)
 				char* url = mr_mprintf("%s://autoconfig.%s/mail/config-v1.1.xml?emailaddress=%s", i==0?"https":"http", param_domain, param_addr_urlencoded); /* Thunderbird may or may not use SSL */
 				param_autoconfig = moz_autoconfigure(mailbox, url, param);
 				free(url);
-				PROGRESS(300+i*50)
+				PROGRESS(300+i*20)
+			}
+		}
+
+		for( i = 0; i <= 1; i++ ) {
+			if( param_autoconfig==NULL ) {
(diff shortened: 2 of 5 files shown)


testbird:
I agree the `emailaddress=` query parameter might make sense to fetch custom login names, even if it might make the provider look stupid for providing credentials with plain text http.

To aleviate the issue of sending the email address to some unrelated server, it could be possible to first probe the http server address without the `emailaddress=` part.

See this crude copy-paste patch for an idea of an improved sequence (but very non-elegant): https://github.com/testbird/deltachat-core/commits/patch-1

r10s:
> To aleviate the issue of sending the email address to some unrelated server, it could be possible to first probe the http server address without the emailaddress= part.

might be that servers, that required the emailaddress parameter reply with "404-Not Found" then. One could easily imagine a server-side script that looks up the given address in a database and throws 404 on failure. so probing won't help, not even with a fake-email.

also, the queried server is not "unrelated" - it is the server hosting the mail service. i am not server expert, but i think before a GET with an fully-qualified url is done, the ip address must be figured out - might be that the full URL is not sent if this fails because the server does not exists. @Ampli-fier do you know?

all in all, we should also keep in mind that this call is done only _once_ the address is configured (not even on every login) and the address - without any other information attached - goes to the _same_ domain delta will connect afterwards. 

there are probably a thousands other placed the email address is more open visible.

also thunderbird (and probably others) do just the same; maybe it's not worth to spend more time on this issue :)

Ampli-fier:
I thought I had seen the change (to try https before http) in a commit already some weeks ago. 

As stated above, the server is not unrelated. The user enters his/her mail domain, e.g. `mail.com`. The attempt to look for the xml will be sent to `autoconfigure.mail.com`, which can be an alias or a DNS entry. This must be resolvable to an IP address. If the user mistyped the domain, the request **can** be sent if by chance the URL can be resolved. Even without autoconfig, the application would ty to send a login request to that (wrong) address anyway. So there is no negative side effect from autoconfigure. If there is no IP resolved, nothing would be sent.

The `emailaddress=` query parameter is to my opinion a bonus provided by the mail service provider. If one mistypes the address name (before the @ sign) the server does not respond with login parameter. If the server would, it would result in an invalid attempt to login. Plus the user would think everything is typed correctly. To me it looks this parameter is mandatory (at least it is for my provider).

testbird:
> If not needed, the response will contain login params. If needed, the autoconfig would fail

Thank you Ampli-fier for testing this with real servers. Sadly, then the idea in
https://github.com/testbird/deltachat-core/blob/a40a7732dec9a1551199bd60d1c1f996bcd9e7d6/src/mrmailbox_configure.c#L508 doesn't seem to work.

The plain text request are probably more irritating than logging it at some providers' web server instead of its (non-existing) config server.
As far as I could try when no autoconfigure server was found with https, there wasn't one reachable through plain html either. So are http-only configuration servers really still relevant these days, or, should they be trusted?

(BTW: There seem commits attached to this pull request that don't seem related to "improve autoconfig".)

testbird:
> also thunderbird (and probably others) do just the same ...

Yea, but on the other hand their "insisting calling home with some user's id data, even if it breaks manual configuration" is a worse example: https://bugzilla.mozilla.org/show_bug.cgi?id=647761

r10s:
> Yea, but on the other hand their "insisting calling home with some user's id data, even if it breaks manual configuration" is a worse example: https://bugzilla.mozilla.org/show_bug.cgi?id=647761

this is not true for delta, however.

if either imap-server, imap-port, smtp-server, smtp-port, smtp-user or server-flags are set, autoconfig is bypassed in delta, see https://github.com/deltachat/deltachat-core/blob/master/src/mrmailbox_configure.c#L488 - the concrete list may change, but the idea is to not to do autoconfig when there is an manual configuration.

testbird:
> this is not true for delta, however.

And it would have surprised me if it would have been true. :-) Thanks for confirming, it doesn't look smart to hard enforce something like this.

r10s:
> (BTW: There seem commits attached to this pull request that don't seem related to "improve autoconfig".)

yes i was on the wrong branch for some time ...

Issue #136, opened 2018-03-28
Subject prefix


r10s:
This issue is about the `Chat:` prefix of all subject lines sent by Delta Chat.

This issue continues the discussion started in at https://github.com/deltachat/deltachat-android/issues/47 at part of other issues.

The idea of the subject prefix `Chat:` was to be equal for all users and make it easier to filter such messages in other e-mail-clients (analogue to `Re:` or `Fwd:`).  
(if other e-mail clients [cannot filter for this](https://github.com/deltachat/deltachat-android/issues/47#issuecomment-376815583), i think these e-mail clients are broken :)

For this point of view, it does not make sense to make it editable to the user or let the user disable the prefix (btw. in one of the next versions the subject will be moved to the encrypted part, same as enigmail does - EDIT: the subject is protected now) 

However, **maybe we can skip the prefix at all?** This could improve the integration with existing MUAs and make Delta Chat messages look more "normal" - for replies we could use the Re: form instead of the text excerpt. This would also make the support of mailing list easier (#125).

EDIT: without the prefix it might also be easier for some MUAs to follow the threads, esp. if Delta adds more correct In-Reply-To and References headers.

testbird:
Using the `Re:` convention seems fine to me. (At least it is common and doesn't introduce yet another thing that is hard to translate.)

testbird:
For filtering of Email-chat compliant messages: What is the `Chat-Version:` header's meaning? Is this the onel?

r10s:
The `Chat-Version` header is for [detecting messenger messages as such](https://github.com/deltachat/spec/blob/master/spec.md#email-chat-specification). If the client allows it, this can be used perfectly for filtering.

testbird:
A thing I didn't realize earlier is that dropping the `Chat:` prefix would also improve the "DC as full MUA" use-case.

testbird:
On the other side, without `Chat:` is there still some way to filter email answers from non-Chat clients into the chat folder on the server (e.g. sieve), before DC gets online and can do it based on the `in-reply-to`?

testbird:
> Edit: This drawback becomes irrelevant if the subject is encrypted anyway.

Actually, the memoryhole standard seems to allow to have different headers in the unencrypted part.

[...]

PR #138, opened 2018-03-29
Write memoryhole

r10s:
moving sensible headers from the unencrypted part to the header of the encrypted payload.

headers moved are: `Subject`, the `Secure-Join`-headers (used for the newQR code verification) and all `Chat-` headers (but  `Chat-Version` as there is a rough idea that this can be used for filtering).

moreover, the `Autocrypt-Gossip` headers are also moved here.

nb: older Delta Chat versions can read these headers already since [v0.11.4](https://github.com/deltachat/deltachat-android/blob/master/CHANGELOG.md#v0114)

information: https://github.com/autocrypt/memoryhole/ (this doc really needs some maintenance)

maybe in the future further headers can be moved to the encrypted part, however, this needs some discussions wirh the memoryhole people; however, at least, the code is now there :)

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -42,6 +42,7 @@ lib_hdr = [
   'mrcontact.h',
   'mrdehtml.h',
   'mrevent.h',
+  'mrerror.h',
   'mrhash.h',
   'mrimap.h',
   'mrjob.h',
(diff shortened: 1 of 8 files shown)


testbird:
I can only comment on leaving the `Chat-Version` header unencrypted, this still does not seem to allow generic filtering classic MUA answers: https://github.com/deltachat/deltachat-core/issues/136#issuecomment-377444585

r10s:
well, the idea is more to leave an option for filtering on the server-side. usually, they can handle headers. i also think some clients can handle this.

PR #140, opened 2018-03-30
support hidden messages

r10s:
hidden messages are needed as the handshake messages for the secure-join messages sent after a QR scan should not appear in the normal chats. Instead, a resulting "system message" as "contact verified" or so should appear.

btw: an alternative to hiding messages would be to allow sending normal MIME-messages without the database background, however, this seems even more complicated as the current approach offers contact_ids, easy deletion and so on. and hiding messages may get handy for other things, too.


testbird:
Do you already have some description of the verification sequence available somewhere, to grasp the concept? Maybe it the user could be asked if he wants to accept the hidden system message (showing the sender and later signing fingerprint?)

PR #141, opened 2018-03-31
allow curl to perform "insecure" SSL connections

Ampli-fier:
allow curl to perform "insecure" SSL connections using flag '--insecure'
Without this flag, curl exited with error 51: The peer's SSL certificate or SSH MD5 fingerprint was not OK.
autoconfig attempts to https URLs did not succeed.

### DIFF
--- a/cmdline/main.c
+++ b/cmdline/main.c
@@ -81,7 +81,7 @@ static uintptr_t receive_event(mrmailbox_t* mailbox, int event, uintptr_t data1,
 			{
 				char* ret = NULL;
 				char* tempFile = mr_get_fine_pathNfilename(mailbox->m_blobdir, "curl.result");
-				char* cmd = mr_mprintf("curl --silent --location --fail %s > %s", (char*)data1, tempFile); /* --location = follow redirects */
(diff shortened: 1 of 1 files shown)


r10s:
i think "insecure" (in fact, probably self-signed) https is better than normal http that would be used otherwise if https fails. so - looks good to me.

r10s:
/me still searching some easeter eggs in the code

Ampli-fier:
```
                       .-^-.
                     .'=^=^='.
                    /=^=^=^=^=\
            .-~-.  :^= HAPPY =^;
          .'~~*~~'.|^ EASTER! ^|
         /~~*~~~*~~\^=^=^=^=^=^:
        :~*~~~*~~~*~;\.-*))`*-,/
        |~~~*~~~*~~|/*  ((*   *'.
        :~*~~~*~~~*|   *))  *   *\
         \~~*~~~*~~| *  ((*   *  /
          `.~~*~~.' \  *))  *  .'
            `~~~`    '-.((*_.-'
```

r10s:
found some eggs!!!!111one1eleven

PR #145, opened 2018-04-06
explaining comment

testbird:
Thanks r10s for explaining (https://github.com/deltachat/deltachat-core/commit/08611fac233109695eed8c990b35875aa146bfbc#commitcomment-28451341)


r10s:
> This state only allows showing incoming messages.

This is not 100% correct. The state says that at some time the contact was validated by a qr code scan.
It's up to the caller what to do with these information.

I think the old explanation is more precise here. But the link mrcontact_is_verfied() is fine,

testbird:
OK, thanks, I think I got you better now.

r10s:
perfect :)

PR #146, opened 2018-04-10
added files and changes to handle uuencoded message parts

csb0730:
This is the first step to include this functionality. Main functions are working but need to be verified.

- New files are mruudecode.c and mruudecode.h.
- Functions are included in mrmimeparser.c

Furthermore to be done:
- Add a full path for extracted decoded data file. In the current implementation simply the current directory is used.
- In source message the uuencoded part is replaced by a hint and name of uuencoded filename. Here a link needs to be placed to access decoded files.

### DIFF
--- a/src/mrmimeparser.c
+++ b/src/mrmimeparser.c
@@ -3,6 +3,7 @@
  *                              Delta Chat Core
  *                      Copyright (C) 2017 Björn Petersen
  *                   Contact: r10s@b44t.com, http://b44t.com
+ *                   Contact: schneider17@gmx.de (cs)
  *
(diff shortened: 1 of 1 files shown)


r10s:
can you please try to add mruudecode.c and mruudecode.h to the PR? this would make discussions easier.

csb0730:
How can I do this? I can't find a possibility :-(
If I try to add the two files here by "Attach files by ... , selecting them, ..." but then I found: These kind of files are not supported! (.c and .h).

Do You have a hint?

csb0730:
I close this PR here because it is replaced by #147.

Issue #90, opened 2018-01-13
uuencoded attachment included in message body is not detected as attachment


csb0730:
If this happens, complete attachment is treated as normal text and is copied to database. The result is that DC freezes if this message will be displayed.

In my case a message size of 15MB occured !

@r10s has some details about it.

csb0730:
@r10s 
beside solving this issue IMHO it would be a good idea to check message size before copying it to database.

Proposal:
  There should be an internal emergency limit for max message size (maybe 100kb) to prevent freeze of the application.
  I suspect that there will be some new cases in future which not will be covered due to misinterpreting a message.

r10s:
> beside solving this issue IMHO it would be a good idea to check message size before copying it to database.

This is definitely an good idea

csb0730:
v0.15.0 (github versioncode=513)

Hi @r10s, 
this happens again now for my phone! DC unusable if certain chat is opened.
Dialog appears that DC is not responding and can be closed by Operating System!

Please introduce as soon as possible an emergency limit for database' message size! Then user at least has the chance to delete the affected message.

Maybe this is related to issue N° 254 from deltachat/deltachat-Android?

r10s:
thank you for the additional information, regarding https://github.com/deltachat/deltachat-android/issues/254 - hard to say if this this related, might me.

csb0730:
It must not but it can be.
But ... really: please introduce an emergency limit for text size (and/or bugfix). A standard user will not have a chance to delete such a message by "backup-search message with external db browser-delete it-reimport backup!" :-/

r10s:
> But ... really: please introduce an emergency limit for text size (and/or bugfix).

yes, I think this is a good idea :)

csb0730:
Hi @testbird,
this issue makes the app unusable if it happens! Not important for v1.0 ?!

r10s:
@csb0730 iiuc the milestone is moved to the project here: https://github.com/orgs/deltachat/projects/2

@testbird thank you for pushing this forward. i have no experience with the github project functionality - maybe we can additionally leave the milestones at the project so that people do not get confused? currently, i also find it easier to have a milestone-tagged list, because when i work on the issue list, it seems to be more focused to me. however, maybe it's only my habit :)
but we should take care not to over-engineer these lists and issues so that new people have a chance to understand what is going on here. keeping things simple seems to be one of the hardest jobs :)

testbird:
[...]

PR #147, opened 2018-04-13
csb0730 uudecode 1 - added files and changes to handle uuencoded message parts (2)

csb0730:
A try to include the changed AND added files in a pull request

### DIFF
--- a/src/mrmimeparser.c
+++ b/src/mrmimeparser.c
@@ -3,6 +3,7 @@
  *                              Delta Chat Core
  *                      Copyright (C) 2017 Björn Petersen
  *                   Contact: r10s@b44t.com, http://b44t.com
+ *                   Contact: schneider17@gmx.com (cs)
  *
  * This program is free software: you can redistribute it and/or modify it under
  * the terms of the GNU General Public License as published by the Free Software
@@ -25,6 +26,8 @@
 #include "mrmimefactory.h"
 #include "mrsimplify.h"
 
+/* new cs */
+#include "mruudecode.h"
 
 /*******************************************************************************
  * debug output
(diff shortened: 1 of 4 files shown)


csb0730:
I seems that into this PR all relevant files are included, so #146 might be closed?

csb0730:
This is the first step to include this functionality. Main functions are working but need to be verified.
- New files are mruudecode.c and mruudecode.h.
- Functions are included in mrmimeparser.c

Furthermore to be done:
- Add a full path for extracted decoded data file. In the current implementation simply the current directory is used.
- In source message the uuencoded part is replaced by a hint and name of uuencoded filename. Here a link needs to be placed to access decoded files.
- Some checks in new functions to secure safe operation of all code parts

src/mrmimeparser.c
```diff
@@ -1024,8 +1027,55 @@ static int mrmimeparser_add_single_part_if_known(mrmimeparser_t* ths, struct mai
 					part = mrmimepart_new();
 					part->m_type = MR_MSG_TEXT;
 					part->m_int_mimetype = mime_type;
-					part->m_msg = simplified_txt;
-					part->m_msg_raw = strndup(decoded_data, decoded_data_bytes);
+					
+                    /* cs:
```
r10s: m_msg needs to be modified because this is displayed to the user, however, m_msg may contain modified data that cannot be uudecoded properly. all data are in m_msg_raw.

thinking it over, I think the best way is to do the uuencode some lines above before mrsimplify_simplify() is called; on decoded_data and decoded_data_bytes then (CAVE: we should check if we can modify the buffers directly).

src/mrmimeparser.c
```diff
@@ -1024,8 +1027,55 @@ static int mrmimeparser_add_single_part_if_known(mrmimeparser_t* ths, struct mai
...
+                        /* any number of parts possible */
+                        uustartpos = mr_find_uuencoded_part(source_str, lineend);
+                        if(uustartpos){
+                            /* todo: path to created files is not handled in the moment */
+                            source_str = mr_handle_uuencoded_part (source_str, uustartpos, lineend);
+                        }
+                        else{
+                            break;
+                        }
+                    }
```
r10s: not sure how the blob is returned can can be given to mrmimepart_new().

src/mruudecode.c
```diff
@@ -0,0 +1,618 @@
+/*******************************************************************************
+ *
+ *                              Delta Chat Core
```
r10s: as mentioned, the header must be equal over all files.

src/mrmimeparser.c
```diff
@@ -3,6 +3,7 @@
  *                              Delta Chat Core
  *                      Copyright (C) 2017 Björn Petersen
  *                   Contact: r10s@b44t.com, http://b44t.com
+ *                   Contact: schneider17@gmx.de (cs)
```
r10s: as mentioned, the header must be equal over all files.

src/mrmimeparser.c
csb0730: Yes, that's true:
 As the new functions are designed to cut out the uuencoded part from the message they should work on *decoded_data*. This means before the message is doubled to *m_msg* and *m_msg_raw*.
 All other functions can stay then as they are if the encoded part is cutted out early enough.

I would prefer to have a link in remaining message text to the attachment if possible. At least a hint with the filename at the cut position. This guarantees the relation to the original text position.

By the way: what means CAVE ?

csb0730: In general it should be possible to modify the buffer directly because the resulting text in each case is shorter or in maximum unchanged if no uuencoded part is found!

src/mrmimeparser.c
csb0730: In the moment only the message without the uuencoded part is finally returned. The decoded data itself is directly written to a file.

src/mrmimeparser.c
csb0730: You may change that?

csb0730:
See my comments above

r10s:
superseeded by https://github.com/deltachat/deltachat-core/issues/90#issuecomment-383955124

PR #169, opened 2018-05-25
encode filenames in 'Content-Disposition: ... filename=...' wrt RFC 2047

### LINKED ISSUE #98: When sending attachments, use proper encoding for filenames
while working on parsing/decrypting messages sent from Delta with [muacrypt](https://muacrypt.readthedocs.io) i noticed that DC encodes filenames for filenames in UTF-8.  As @dkg pointed out on #autocrypt this rather needs to use a special encoding used in headers (header values need to be 7-bit clean).  For example, mutt creates this for an umlaut-containing filename:

> Content-Disposition: attachment; filename*=iso-8859-1''%FCbersicht%2Etxt

enigmail rather uses encodings as found in [RFC2047](https://tools.ietf.org/html/rfc2047) and particularly the "8. example" section. 

Mutt seems to be using [RFC5987](https://www.rfc-editor.org/rfc/rfc5987.txt) which [some claim](https://stackoverflow.com/a/216777) is the right solution at least in http headers.  @dkg says the mutt filename parameter encoding is rather what RFC2231 specifies ...

r10s:
should fix #98


r10s:
okay, wrt to the discussion at #98, these changes are correct, but not sufficient.

PR #172, opened 2018-06-02
initialize from_addr_norm with NULL

labdsf:
Otherwise, if mimeparser is NULL, "goto cleanup" is triggered and
uninitialized pointer will be freed.

clang warns about this during compilation.


r10s:
good catch, thanks :)

Issue #47, opened 2017-10-02
New Name for the "Chats" folder


r10s:
Currently, we move messages sent by Delta Chat and replies to them to the special IMAP-folder "Chats". If this folder does not exist, it is created by Delta Chat (yes, there are [some bugs](https://github.com/deltachat/deltachat-android/issues/108) with that, but this is the plan how it should work :)

However, the name "Chats" was not a good choice as it is reserved by some providers, Google among them. 

So, currently, we're searching for a new generic name for this folder.

- the name should be unique
- it should be clear what gets in there
- it should not be reserved, so. eg. `Chat` may be a bad choice as well
- there is no need to bind the name to "Delta Chat", however, it is not impossible (maybe even the best choice) (we avoid this binding also for other generic things, eg. the `Chat:` prefix in the subject or the [internal header names](https://delta.chat/en/spec))  
- it is not planned to use customized names for the user (different names may make things even more complicated in the future)

Suggestions up to now:

`DeltaChats`

`DeltaChat`

`IMAPChat`

testbird:
Hmm, compared to the other common default folders like Sent, Junk, Draft, Spam, Trash, ..., the content does not consist of formal E-Mail Letters...

So other ideas: Messages, Conversation, ...?

testbird:
Messenger
delta.chat  <-- actually hinting to the URL if anybody is not in the know.

r10s:
> the content does not consist of formal E-Mail Letters...

Well, replies from normal MUAs to chats started in Delta will be moved there, too. However, if you mean the typically missing "Dear madam/sir ... Best regards ..." - that's true.

STPKITT:
Blank in the name are not allowed?

The rest of the decision is highly dependend on which kind of people you're targeting. Experts/Nerds etc. will always comprehend what that new folder in their inbox is all about no matter what name it has, BUT every other person not into tech will not.
So I highly recommend that "Delta" is part of the name, I personally would vote for DeltaChat.

r10s:
> Blanks in the name are not allowed?

Not sure about what the spec says. But I'm quite sure, there will be at least one app or one agent having problem with spaces in foldernames. It may also be more difficult for the user, eg. when adding a filter or whatever. 
So, maybe it is a good idea not to use spaces, event if the standard says that we can use some and even if Delta has no problems with them.

STPKITT:
Ok, I agree. No blanks often means less trouble.

Hocceruser:
[...]

PR #173, opened 2018-06-02
Use INBOX.Chats as fallback if Chats can't be created

t-oster:
This fixes https://github.com/deltachat/deltachat-android/issues/103 and probably other providers (maybe https://github.com/deltachat/deltachat-android/issues/28).

If it is not possible to create the folder `Chats` we also try `INBOX.Chats` and use this if creation succeeds. I tested this on a fresh domain factory account and it works.


r10s:
looks very good, thank you very much for pushing this forward.

i think we should also make a decision to #47

t-oster:
Thanks for merging. If I understand correctly the separator is not always '.' and could also be '\' but I will look that up and maybe try to determine the separator while iterating trough all folders.I will make a new PR then

r10s:
sounds very good. as far as i know, there is no "separator" in that sens in the imap folder names at all, so, in fact, there may be different conventions (but i am really not experiences in this stuff) (just edited your comment to make the backslash visible (wondering what there is between the quotes :))

PR #175, opened 2018-06-04
Feature: Detect IMAP separator for INBOX subfolder

t-oster:
Instead of using hardcoded `INBOX.Chats` as fallback if `Chats` can't be created (#173) we detect the IMAP folder-delimiter as part of the list folders process and use this to construct the fallback folder name.

Please review if I did everything correct with C strings, because I am not that familiar with C and don't want to introduce security holes.


r10s:
looks good to me.

PR #176, opened 2018-06-04
Workaround for hMail. Fixes deltachat/deltachat-android#67

t-oster:
This commit retries the SMTP login if the method is PLAIN and it failed. See deltachat/deltachat-android#67 for details.

### DIFF
--- a/src/mrsmtp.c
+++ b/src/mrsmtp.c
@@ -19,7 +19,7 @@
  *
  ******************************************************************************/
 
-
+#include <unistd.h>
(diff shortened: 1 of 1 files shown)


r10s:
i have not tested this out with a hMailserver, but logically, it looks good to me.
very good that the fix is only executed if connection fails before, so existing stuff should not break. 

the fix seems to be in-line with the research from @Ampli-fier and other in https://github.com/deltachat/deltachat-android/issues/67 so there seems to be good chances that this will work. merging :)

PR #179, opened 2018-06-04
Rename Chats folder to DeltaChats. Fixes #47

### LINKED ISSUE #47: New Name for the "Chats" folder

t-oster:
This renames the Chats folder to DeltaChats and thus should fix #47 and some other issues (gmail, yahoo).
IMHO it makes the function somewhat hard to read, so feel free to refactor before merging.
I just did a quick test on my domainfactory account:
I just upgraded my existing Android install and after I sent the first message, it successfully renamed the old Chats folder to DeltaChats (under Inbox).
However, now I have a dead Chats folder in Thunderbird which cannot be deleted, because it does not exist ;).

### DIFF
--- a/src/mrchat-private.h
+++ b/src/mrchat-private.h
@@ -61,7 +61,8 @@ int             mrchat_are_all_members_verified__ (mrchat_t*);
 
 
 #define         MR_CHAT_PREFIX              "Chat:"      /* you MUST NOT modify this or the following strings */
-#define         MR_CHATS_FOLDER             "Chats"      /* if we want to support Gma'l-labels - "Chats" is a reserved word for Gma'l */
+#define         MR_CHATS_FOLDER             "DeltaChats" /* Folder where the Chats will be moved to */
+#define         MR_CHATS_FOLDER_LEGACY      "Chats"      /* Folder name in old version. Will be renamed to new folder if present */
 
 
 #ifdef __cplusplus
--- a/src/mrimap.c
+++ b/src/mrimap.c
@@ -302,7 +302,7 @@ static int init_chat_folders__(mrimap_t* ths)
 	int        success = 0;
 	clist*     folder_list = NULL;
 	clistiter* iter1;
-	char       *normal_folder = NULL, *sent_folder = NULL, *chats_folder = NULL;
+	char       *normal_folder = NULL, *sent_folder = NULL, *chats_folder = NULL, *chats_folder_legacy = NULL;
 
 	if( ths==NULL || ths->m_hEtpan==NULL ) {
 		goto cleanup;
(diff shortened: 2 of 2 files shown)


t-oster:
sorry I accidentally formatted everything... will force-push a cleaned version in a minute

t-oster:
here we go

src/mrchat-private.h
```diff
@@ -61,7 +61,8 @@ int             mrchat_are_all_members_verified__ (mrchat_t*);
 
 
 #define         MR_CHAT_PREFIX              "Chat:"      /* you MUST NOT modify this or the following strings */
-#define         MR_CHATS_FOLDER             "Chats"      /* if we want to support Gma'l-labels - "Chats" is a reserved word for Gma'l */
+#define         MR_CHATS_FOLDER             "DeltaChats" /* Folder where the Chats will be moved to */
```
r10s: i think the name should be `DeltaChat` (without trailing `s`) and not `DeltaChats`

src/mrchat-private.h
```diff
@@ -61,7 +61,8 @@ int             mrchat_are_all_members_verified__ (mrchat_t*);
...
+#define         MR_CHATS_FOLDER_LEGACY      "Chats"      /* Folder name in old version. Will be renamed to new folder if present */
```
r10s: i would also be fine in not converting the legacy folder:
- avoid additional levels of complexity and testing
- future changes may be more complicated
- it's no breaking change, just a new folder the legacy folder stays there and the user can move messages if wanted.
- Delta Chat is still beta

t-oster:
Totally agree. People could manually rename if wanted

t-oster:
Then its just a matter of changing the constant, I think you dont need a PR for that

r10s:
>  I think you dont need a PR for that

you're right, seems not to be needed. in fact, i was not really aware that it is only a single constant, sorry for the additional effort.

PR #181, opened 2018-06-05
Fallback to Copy/Delete if Move not possible

t-oster:
The mailserver of @Ji-eF does not work with the used MOVE command, so messages are left in the INBOX (see deltachat/deltachat-android#307).
I found the hint in the source that many mailservers don't support the MOVE operation so we should fallback to COPY/DELETE.
This is my first attempt to do that. The copy part works, however the delete part does not.

EDITED. See below

The String from the test-server is:
```
IMAP-Capabilities: IMAP4rev1 LITERAL+ SASL-IR LOGIN-REFERRALS ID ENABLE SORT SORT=DISPLAY THREAD=REFERENCES THREAD=REFS MULTIAPPEND UNSELECT IDLE CHILDREN NAMESPACE UIDPLUS LIST-EXTENDED I18NLEVEL=1 CONDSTORE QRESYNC ESEARCH ESORT SEARCHRES WITHIN CONTEXT=SEARCH LIST-STATUS QUOTA

### DIFF
--- a/src/mrimap.c
+++ b/src/mrimap.c
@@ -1708,9 +1708,23 @@ int mrimap_markseen_msg(mrimap_t* ths, const char* folder, uint32_t server_uid, 
 				struct mailimap_set* res_setdest = NULL;
 				r = mailimap_uidplus_uid_move(ths->m_hEtpan, set, ths->m_moveto_folder, &res_uid, &res_setsrc, &res_setdest); /* the correct folder is already selected in add_flag__() above */
 				if( is_error(ths, r) ) {
-					mrmailbox_log_info(ths->m_mailbox, 0, "Cannot move message.");
-					goto cleanup;
-				}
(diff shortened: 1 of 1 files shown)


t-oster:
Since I would need the `rfc724_mid` for using `mrimap_delete_msg` I just copied the relevant parts to delete the message. However I am not sure this works and it is correct that way.

r10s:
at a first glace, this looks good to me.

r10s:
> Do we have to update any IDs in the DC database, because the copy is a new message and the original one is deleted?

yes, but this is already done by the caller (it is also needed for move)

> This is my first attempt to do that. The copy part works, however the delete part does not.

Does it work for you in the edited version? the code looks as it should work :)

t-oster:
Yes, it works on the Server of @Ji-eF. However Messages which are from an non-DC initiated chat are not moved (part of https://github.com/deltachat/deltachat-android/issues/307). You can merge this and I will have a look into the other problem

t-oster:
(there is no other problem. This fixes https://github.com/deltachat/deltachat-android/issues/307

PR #185, opened 2018-06-10
Use bundeled libetpan by default

flub:
Using the system-libetpan causes pain to people who have out-dated
library versions, additionally it is nice to by default create a
library which uses dependencies as close as possible to the android
version for testing purposes.  This makes sure that the bundeled
sub-projects are used by default.

Additionally it makes sure to not install these bundeled dependencies
and only statically link the libdeltachat.so against them.  This is
beneficial to not conflict with any system-installed versions of these
libraries, especially since some bundled libraries are stripped down
and would not provide indirect users of the library the full library
functionality.

It does still provide an option to use the system-provided libetpan if
desired however, this makes sure libdeltachat plays nice with normal
package managers etc.


t-oster:
Hi, please see #186

t-oster:
However if I use your master branch, it works. So there must be mismatches between commit https://github.com/deltachat/deltachat-core/commit/46513f194250026c645a07f569055df97f9b9275 and your PR

t-oster:
And if I pull your master into current master, it fixes #186 so this PR should be merged IMHO in order to fix #186 introduced by 46513f1

flub:
Thanks for confirming this fixed the build, I've merged this now.

Issue #36, opened 2017-07-04
enigmail key export format is incompatible to import in delta chat


haeckser:
I had a crash on upgrade to 0.92 and the program started from scratch (as if the DB was initialized with a new one). I would like to reimport my keys which I had imported into Thunderbird with enigmail.
The key management (in thunderbird) offers the option to export public and private key, but this format cannot be used for import in Delta Chat
-> additional import format (in general for existing keys) to allow transfer from enigmail
->documentation improvement to give a clear statement that keys generated in Delta Chat should be exported AND stored somewhere. 
I disabled encryption in the first shot in 0.9.2m because I planned to reimport the keys.
But I could only import the public key back from Thunderbird. And so I enabled encryption again with the intention to generate a new key and the hope that it will be published automatically and will overwrite the old public key at contacts automatically. But it seems that the missing imported  private key  (info about tells me 1 public and 0 private keys) blocks encryption. No clear message or signal for the user to understand this behaviour.

haeckser:
I see in code, that the import function is searching for the start target, localize the corresponding end target, delete it and import the whole block. Enigmail is exporting BOTH key in one file and so block is the first key + start. second key and end mark of second key -> garbage
should search for the end tag and import block between and retry search from end tag of first key

r10s:
can you confirm that this is still true for the current version?
and does it only affect enigmail or just gpg? a concrete advice how to generate a key that cannot be imported in delta would also be helpful.

Thanks in advance :)

r10s:
> I see in code, that the import function is searching for the start target, localize the corresponding end target, delete it and import the whole block. Enigmail is exporting BOTH key in one file and so block is the first key + start. second key and end mark of second key -> garbage
should search for the end tag and import block between and retry search from end tag of first key

Thank you fro the very detailed information, this issue should be fixed by the new armored parser now which is also used for key importing:
https://github.com/deltachat/deltachat-core/commit/0758641ef7b2c61d501251d21a96ca9037880e0f

The commit mentioned above checks that the second block/key/whatever is really ignored.

I think this issue can be closed now, if the problem still exists, please reopen.

PR #191, opened 2018-06-15
Fix: Import keys where public and private key are in the same file

t-oster:
When I export my GPG Key via Thunderbird/Enigmail (after I removed the passphrase), it creates a single .asc file which contains the public and private key.
This patch allows DC to also import such files. Since it only comes into action when the file starts with a public key, this should not break anything, eventhough it might not work in all cases.
Related to #36 (which is IMHO not fixed) and https://github.com/deltachat/deltachat-android/issues/98

### DIFF
--- a/src/mrmailbox_imex.c
+++ b/src/mrmailbox_imex.c
@@ -786,7 +786,6 @@ static int import_self_keys(mrmailbox_t* mailbox, const char* dir_name)
 	if( mailbox==NULL || mailbox->m_magic != MR_MAILBOX_MAGIC || dir_name==NULL ) {
 		goto cleanup;
 	}
-
 	if( (dir_handle=opendir(dir_name))==NULL ) {
 		mrmailbox_log_error(mailbox, 0, "Import: Cannot open directory \"%s\".", dir_name);
 		goto cleanup;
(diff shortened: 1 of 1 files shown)


src/mrmailbox_imex.c
```diff
@@ -815,7 +814,17 @@ static int import_self_keys(mrmailbox_t* mailbox, const char* dir_name)
...
+			 * However some programs (Thunderbird/Enigmail) put public and private key
+			 * in the same file, so we check if there is a private key following */
+			char* following_private_key = strstr(buf, "-----BEGIN PGP PRIVATE KEY BLOCK");
+			if (following_private_key == NULL) {
+				continue; /* this is no error but quite normal as we always export the public keys together with the private ones */
+			}
+			/*
+			 * If so, we point the buffer to that private key
+			 */
+			buf = following_private_key;
```
r10s: this will cause problems when free() is called for `buf` afterwards (`following_private_key` is just a pointer inside `buf` and is not a pointer returned by malloc() and cannot be free()'d).

the solution may be to copy buf using memcpy() or to use an additional const-pointer key_start or so that is set up but not free()'d.

maybe we can use a generic `const char* private_key` that is equal to buf by default but may be modified and is never free()'d.

r10s:
okay, i do :)

r10s:
now it looks good to me - @t-oster does it still work for your as you intended?

PR #206, opened 2018-07-05
context requests -> contact requests

Ampli-fier:
Corrected Info message from "context requests" to "contact requests"


r10s:
good catch :)

PR #208, opened 2018-07-11
5th rework of uudecoding functions - renaming all to "dc_" naming convention

csb0730:
@r10s: Ready for testing now.

### DIFF
--- a/src/dc_uudecode.c
+++ b/src/dc_uudecode.c
@@ -558,18 +558,17 @@ int dc_uudecode(char** ret_binary, size_t uudecoding_buffer_len, const char* uu_
 	char* p;
 	char  ch;
 	int   n;
-	int   backquote_found = 0;
+	int   backquote_found   = 0;
 	int   ret_decoded_bytes = 0;
 
 
-	/* open stream for output*/
-	FILE *fp = open_memstream(ret_binary, &uudecoding_buffer_len);
-
+    char* p_binary_buffer   = *ret_binary;  // initialize write buffer pointer
+    
 
 	/**
 	 *  decoding line by line
 	 */
-	while( (line_len = dc_getline (&line, nextlinestart, &nextlinestart)) ){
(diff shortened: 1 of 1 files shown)


src/dc_uudecode.c
```diff
@@ -32,31 +104,597 @@
...
+	// line decoding
+	char* p;
+	char  ch;
+	int   n;
+	int   backquote_found = 0;
+	int   ret_decoded_bytes = 0;
+
+
+	/* open stream for output*/
+	FILE *fp = open_memstream(ret_binary, &uudecoding_buffer_len);
```
r10s: this seems not to be available under Android, the patch does not compile

csb0730: This is unfortunate. Here I used gcc at debian to develop all code and a bigger issue now :-/

I need to replace all the stream related functions with individual ones.

Ok, I'll see what I can do and suspect that this needs now a few days of work ;-)

Coming back if ready ...

csb0730: I did a little research about alternatives and found _fmemopen()_. First I didn't want to use it because it is recommended for reading operations in standard examples. But it can be used for writing to buffers too. 

Is that function available at Android? If yes, that would be an alternative :-)

r10s: i think, these functions are mainly for feeding functions _only_ accepting a FILE handle which seems not to be needed in this case.
we should just access the given memory directly.

i've also added a comment to the belonging issue - i am still not convinced that uuencode is needed at all ... https://github.com/deltachat/deltachat-core/issues/90#issuecomment-404455605

csb0730: Yes You're right. We can access the memory directly. Maybe this is at the end more easy and better than I thought at the beginning :-).
For me it was only the quicker way to choose the method in using open_memstream() because the examples I had worked with real files.

I'll have a solution ;-)

csb0730:
Easier than thought. See new PR.

PR #209, opened 2018-07-11
Rework failed jobs

r10s:


r10s:
this targets https://github.com/deltachat/deltachat-android/issues/102

PR #227, opened 2018-07-28
Fix typos

ralphtheninja:


r10s:
thanks for the corrections -  
seems is if i tend to leave out the vowels at the word endngs :)

PR #236, opened 2018-08-14
Replace symlinks with real files

ralphtheninja:
This replaces all the symlinks inside `libs/libetpan/include/libetpan` with the files pointed to. We need this for two reasons:

* symlinks doesn't work on windows
* `npm publish` removes symlinks (for the above reason) so system wide installation of `libetpan-dev` is required for dependent modules

I'd prefer it if we could fix this some other way, i.e. by changing the include paths given to the compiler (via `-I`). But I couldn't get it to work.

Also see https://github.com/deltachat/deltachat-node/issues/66

cc @r10s @Jikstra


r10s:
in https://github.com/deltachat/deltachat-core/commit/e70475f34e06a3d871a57b96da95e3f28212d8af , i've removed the `-a` from the copy command so that update-libetpan.sh creates real files instead of symlinks (see https://github.com/deltachat/deltachat-cli/issues/35#issuecomment-412833951 ).

in https://github.com/deltachat/deltachat-core/commit/5d3f7ac28f352dcd32cd3cf1d562e8d5d198d3b3 i've called update-libetpan.sh then which creates roughly the same result as this pr that can be closed therefore.

PR #250, opened 2018-08-26
Enable Travis

ralphtheninja:
This makes sure to install everything we need and then builds the library and gives a stepping stone to start implementing tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,30 @@
+sudo: required
+
+addons:
+  apt:
+    sources:
+    - ubuntu-toolchain-r-test
+    packages:
+    - python3-pip
+    - python3.5
+    - g++-7
+
+language: c
+
+before_install:
+  - export CC=gcc-7
+  - export CXX=g++-7
+  - sudo ln -sf /usr/bin/python3.5 /usr/bin/python3
+  - sudo pip3 install meson
+  - wget https://github.com/ninja-build/ninja/releases/download/v1.8.2/ninja-linux.zip
+  - unzip ninja-linux.zip
+  - sudo cp ninja /usr/bin
+  - wget http://http.debian.net/debian/pool/main/c/cyrus-sasl2/cyrus-sasl2_2.1.27~101-g0780600+dfsg.orig.tar.xz
+  - tar xfv cyrus-sasl2_2.1.27~101-g0780600+dfsg.orig.tar.xz
+  - pushd cyrus-sasl2-2.1.27~101-g0780600+dfsg.orig && ./autogen.sh && make && sudo make install && popd
+
+install:
+  - sudo apt install -y libssl-dev libsqlite3-dev libbz2-dev zlib1g-dev
+
+script:
+  - mkdir -p builddir && cd builddir && meson && ninja -v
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,30 @@
...
+  - unzip ninja-linux.zip
+  - sudo cp ninja /usr/bin
+  - wget http://http.debian.net/debian/pool/main/c/cyrus-sasl2/cyrus-sasl2_2.1.27~101-g0780600+dfsg.orig.tar.xz
+  - tar xfv cyrus-sasl2_2.1.27~101-g0780600+dfsg.orig.tar.xz
+  - pushd cyrus-sasl2-2.1.27~101-g0780600+dfsg.orig && ./autogen.sh && make && sudo make install && popd
+
+install:
+  - sudo apt install -y libssl-dev libsqlite3-dev libbz2-dev zlib1g-dev
+
+script:
```
ralphtheninja: @r10 WDYT about adding a `Makefile` in the root of the project so we can put common commands in there, e.g. `make build`, `make test`, `make install`, `make uninstall` etc. This would also allow us to simplify `README.md` and we can instead refer to make commands.

.travis.yml
```diff
@@ -0,0 +1,30 @@
...
+  - mkdir -p builddir && cd builddir && meson && ninja -v
```
ralphtheninja: This would ideally be `make build` and later on when we have tests implemented `make build && make test`.

.travis.yml
```diff
@@ -0,0 +1,30 @@
...
+  - sudo pip3 install meson
+  - wget https://github.com/ninja-build/ninja/releases/download/v1.8.2/ninja-linux.zip
+  - unzip ninja-linux.zip
+  - sudo cp ninja /usr/bin
+  - wget http://http.debian.net/debian/pool/main/c/cyrus-sasl2/cyrus-sasl2_2.1.27~101-g0780600+dfsg.orig.tar.xz
+  - tar xfv cyrus-sasl2_2.1.27~101-g0780600+dfsg.orig.tar.xz
+  - pushd cyrus-sasl2-2.1.27~101-g0780600+dfsg.orig && ./autogen.sh && make && sudo make install && popd
+
+install:
+  - sudo apt install -y libssl-dev libsqlite3-dev libbz2-dev zlib1g-dev
```
ralphtheninja: Installing `ninja-build` above since installing it via `apt` yields a too old version.

ralphtheninja: Also installing `meson` before this step since installing via `apt` fails (no such package).

.travis.yml
r10s: this makes sense, however i have no idea how this relates to meson/ninja (cc @flub)

r10s:
lgtm (but this is not worth much, do not have much experience with travis or makefiles :)
however, as this does not affect other parts, just give it a try )

@ralphtheninja awesome you push this forward :)

.travis.yml
ralphtheninja: Well `make build` would essentially do `mkdir -p builddir && cd builddir && meson && ninja`. It's merely a refactoring step as I see it.

PR #258, opened 2018-09-04
some initial test code, also fixing python callback machinery to actually ca…

hpk42:
…ll the callback (thanks @flub for the hint)

note that this adds a "--user" and "--password" option to the pytest invocation. The options need to specify credentials for acecssing a working smtp/imap account.  If you don't specify it, according tests are skipped. 

Problems:

- [x] how to systematically get at event types (100 = INFO event, etc) as specified in the macro section of https://deltachat.github.io/api/group__DC__EVENT.html#ga2e4cc3e6e1c3ba8f152b2cf94632a967
- [x] the 100 INFO event should have a string in the "data2" argument but it's an integer -- that's a problem of the bindings, no?


r10s:
wrt the callback-parameters: just some ideas:

- in the java-bindings, we take the simple approach of treating `data1` and `data2` always as integers in the first place when [calling from c-land into java-land](https://github.com/deltachat/deltachat-android-ii/blob/master/jni/dc_wrapper.c#L169). if we're then in java-land later and we know by the event type that it is a string, we have a [simple wrapper-function](https://github.com/deltachat/deltachat-android-ii/blob/master/jni/dc_wrapper.c#L1259) that takes `data1` or `data2` and returns a String.

- in the node-bindungs @ralphtheninja and @Jikstra create the "correct" JavaScript object [already in c-land](https://github.com/deltachat/deltachat-node/blob/master/src/module.c#L52)

- might also be an idea to write a python-helper-function to expand the callback to `callback(int event, uintptr_t data1, const char* data1str, uintptr_t data2, const char* data2str)` so that maybe the python-cffi-generator has the correct types. don't know if this can work.

btw. to find out if an event takes a string, it is safe to "know" it by the type (as for DC_EVENT_INFO). however, for generic approaches or wrappers, DC_EVENT_DATAx_IS_STRING(event) may also be helpful.

btw#2. for napi, we did a c-class to get the events synchronously, this might also be useful for testing, see https://github.com/deltachat/deltachat-node/blob/master/src/eventqueue.c

python/src/deltachat/__init__.py
```diff
@@ -11,4 +11,8 @@ def py_dc_callback(ctx, evt, data1, data2):
     CFFI only allows us to set one global event handler, so this one
     looks up the correct event handler for the given context.
     """
-    return _DC_CALLBACK_MAP.get(ctx, lambda *a: 0)
+    callback = _DC_CALLBACK_MAP.get(ctx, lambda *a: 0)
+    ret = callback(ctx, evt, data1, data2)
```
flub: How about:
```
try:
   ret = callback(ctx, evt, data1, data2)
except:  # I think a bare except is appropriate here
    ret = 0
```

python/src/deltachat/__init__.py
```diff
@@ -11,4 +11,8 @@ def py_dc_callback(ctx, evt, data1, data2):
...
+    if ret is None:
```
flub: probably `if not isinstance(ret, int0):`

python/tests/test_lowlevel.py
```diff
@@ -0,0 +1,40 @@
+from __future__ import print_function
```
flub: why still bother supporting python 2.7?

python/tests/conftest.py
```diff
@@ -13,3 +15,56 @@ def register_dc_callback(monkeypatch):
...
+                                  capi.ffi.NULL, capi.ffi.NULL)
+    yield ctx
+    capi.lib.dc_close(ctx)
+
+
+@pytest.fixture
+def dc_threads(dc_context):
+    quitflag = threading.Event()
+    t1 = threading.Thread(target=imap_thread, name="imap", args=[dc_context, quitflag])
+    t1.setDaemon(1)
```
flub: Since these threads intend to be cleaned up, why not be strict and not have them be daemon threads and join them before the test finishes?  It's slower but once there's many tests it'll help I think.  We can then add pytest-timeout to debug any hangs

flub:
For the types of the data arguments, I think you can do `ffi.string(ffi.cast('char*', data2))` to get the string (untested)

python/tests/test_lowlevel.py
hpk42: because i use it ;)

python/tests/test_lowlevel.py
```diff
@@ -3,7 +3,10 @@
 import re
 from deltachat import capi
 from deltachat.capi import ffi
-import queue
+try:
+    from Queue import queue
+except ImportError:
+    from queue import Queue
```
ralphtheninja: Wat? :)

r10s: why not also try `from queue import queue` and from `Queue import Queue` :)

python/tests/test_lowlevel.py
flub: <guess the followup question> ;)  no really.  what's there to gain in this case?

python/src/deltachat/__init__.py
```diff
@@ -11,4 +13,38 @@ def py_dc_callback(ctx, evt, data1, data2):
...
+        data2 = ffi.string(ffi.cast('char*', data2))
+    evt_name = get_dc_event_name(evt)
+    try:
+        ret = callback(ctx, evt_name, data1, data2)
+        if event_sig_types & 4:
+            return ffi.cast('uintptr_t', ret)
+        elif event_sig_types & 8:
+            return ffi.cast('int', ret)
+    except:
+        raise
```
flub: i'm assuming this is for debugging purposes only

python/src/deltachat/__init__.py
```diff
@@ -11,4 +13,38 @@ def py_dc_callback(ctx, evt, data1, data2):
...
+
+def set_context_callback(dc_context, func):
+    _DC_CALLBACK_MAP[dc_context] = func
+
+
+def clear_context_callback(dc_context):
+    _DC_CALLBACK_MAP.pop(dc_context, None)
+
+
+def get_dc_event_name(integer, _DC_EVENTNAME_MAP={}):
```
flub: cut but scary.  why not make `_DC_EVENTNAME_MAP` a module-level constant?  It's a bit more conventional

python/src/deltachat/account.py
```diff
@@ -0,0 +1,106 @@
...
+class Account:
+    def __init__(self, db_path, logcallback=None):
+        self.dc_context = ctx = capi.lib.dc_context_new(
+                                  capi.lib.py_dc_callback,
+                                  capi.ffi.NULL, capi.ffi.NULL)
+        capi.lib.dc_open(ctx, db_path, capi.ffi.NULL)
+        self._logcallback = logcallback
+
+    def set_config(self, **kwargs):
+        for name, value in kwargs.items():
```
flub: will all names always be valid python identifiers?  The `**kwargs` would stop deltachat from using e.g. `account-name` as a config name.

python/src/deltachat/account.py
```diff
@@ -0,0 +1,106 @@
...
+import threading
+import requests
+from . import capi
+import deltachat
+from .capi import ffi
+
+
+class Account:
+    def __init__(self, db_path, logcallback=None):
+        self.dc_context = ctx = capi.lib.dc_context_new(
```
flub: `self.dc_context` is low-level.  I'd consider making it private

python/src/deltachat/account.py
```diff
@@ -0,0 +1,106 @@
...
+        # also can't now close the database because the
+        # threads might still need it
+        # capi.lib.dc_close(self.dc_context)
+
+    def process_event(self, ctx, evt_name, data1, data2):
+        assert ctx == self.dc_context
+        if self._logcallback is not None:
+            self._logcallback((evt_name, data1, data2))
+        callname = evt_name[3:].lower()
+        method = getattr(self, callname, None)
```
flub: I'm not a fan of the lookup on self here.  Dependency injection being better than inheritance and all that.

python/src/deltachat/account.py
```diff
@@ -0,0 +1,106 @@
...
+        if method is not None:
+            return method(data1, data2) or 0
+        # print ("dropping event: no handler for", evt_name)
```
flub: is `self._logcallback` suitable for this?  Or not?

python/tests/test_lowlevel.py
flub: huh, github ate half of my response.  anyway, I'd be very careful about still supporting 2.7 at this point to be honest.

python/tests/test_lowlevel.py
flub: all because @hpk42 wants to support an ancient version of Python ;)

PR #260, opened 2018-09-05
Update update.sh

devSJR:
missing word?


docs/update.sh
```diff
@@ -7,7 +7,7 @@ doxygen
 
 
 # HACK: we use the doxygen commands @defgroup and @addtogroup to sort #define's,
-# however,  doxygen files them under the menu entry "Modules".
```
r10s: i think the sentence is correct. "files" are a verb here (to file), not a noun (the files).

Issue #268, opened 2018-09-08
expose is_smtp_connected|is_imap_connected


hpk42:
While hacking on the python bindings i noticed that i'd like to call ``is_smtp|imap_connected`` but this is not exposed in deltachat.h and it uses ``dc_smtp|imap_t`` structs which are also not exposed if i see it correctly.  Could we get a top-level API that just takes a ``dc_context_t*``?  I think there is a 1:1 reference relation between context and smtp|imap structs so it doesn't semantically change much.

If changing the signature of the two is_*_connected functions is not workable, i am also fine with new functions ("dc_get_smtp_status" -> 0: no connection, 1: connection, 2: auth/fully-operable) and the respective imap parts.  

grateful for a PR either way as i want to continue writing a "setup two live accounts, send and receive messages between them" functional test.

r10s:
i would leave the `dc_smtp_is_connected(dc_smtp_t*)` and `dc_imap_is_connected(dc_imap_t*)` as they are now - `dc_imap_t` and `dc_smtp_t` are independent from `dc_context_t` (there are only dependencies for logging), so, these function have their purpose.

but we can easily add `dc_is_smtp_connected(dc_context_t*)` and a  `dc_is_imap_connected(dc_context_t*)` functions. however, the suggested tri-state is not easily doable - if a connection did not get auth/fully-operable state, it is not treated as being connected. if the connection is lost, everything is lost and a complete reconnect is done.

so, i would just forward the values from `dc_smtp_is_connected(dc_smtp_t*)` and `dc_imap_is_connected(dc_imap_t*)`.

@hpk42 doing this now.

hpk42:
Hum, after reading your PR's docstrings i am wondering if it's a good idea to have these top level funcitons.
Going to try rather figuring things out from events, i think.  But I'd like to have clearer events like "DC_EVENT_IMAP_CONNECTED" "DC_EVENT_IMAP_DISCONNECTED" (and the same for smtp). with NULL DATA1/DATA2 for now.  Is that as easily possible as was #PR269 ?

r10s:
i can easily send DC_EVENT_IMAP_CONNECTED and DC_EVENT_SMTP_CONNECTED, however, for sending DC_EVENT_IMAP_DISCONNECTED, there is no nice logic.

however, in practice, a DISCONNECT is always followed directly by the attempt to reconnect.

maybe we can send the state of a connect/reconnect attempt just by an event.
[...]

PR #269, opened 2018-09-08
add high-level functions to check the imap- and smtp-connection

r10s:
introduce two function that expose the internal imap-connection and smtp-connection state to the dc_context_t object.

see #268

### DIFF
--- a/src/deltachat.h
+++ b/src/deltachat.h
@@ -241,6 +241,8 @@ void            dc_openssl_init_not_required (void);
 // connect
 void            dc_configure                 (dc_context_t*);
 int             dc_is_configured             (const dc_context_t*);
+int             dc_is_imap_connected         (const dc_context_t*);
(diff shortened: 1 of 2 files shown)


src/deltachat.h
```diff
@@ -241,6 +241,8 @@ void            dc_openssl_init_not_required (void);
 // connect
 void            dc_configure                 (dc_context_t*);
 int             dc_is_configured             (const dc_context_t*);
+int             dc_is_imap_connected         (const dc_context_t*);
+int             dc_is_smtp_connected         (const dc_context_t*);
```
hpk42: is it a good idea to have the same name for a function that is defined differently in dc_configure.c and dc_smtp.c etc?

flub: I'm pretty happy with this naming btw, seems logical.  but let's see in the issue if this is PR is still desired I guess

PR #270, opened 2018-09-08
send an event if imap/smtp state changes

r10s:
alternative to pr #269

would comment this properly if accepted :)


flub:
Maybe it makes sense to add a state diagram of smtp and imap if their "state" is being exposed.  I certainly don't know if the current connected/not connected exposure is more or less fine or not

r10s:
i think we should close this pr and go for a more simple DC_EVENT_IMAP_CONNECTED / DC_EVENT_IMAP_DISCONNECTED

PR #271, opened 2018-09-08
send an event if imap/smtp connection gets established

r10s:
3rd alternative to #270 and #269 which i will close just now to avoid too much confusion.

corresponding issue: #268

would comment this properly if accepted :)

### DIFF
--- a/src/dc_imap.c
+++ b/src/dc_imap.c
@@ -1155,6 +1155,7 @@ static int setup_handle_if_needed(dc_imap_t* imap)
 
 	dc_log_info(imap->context, 0, "IMAP-login as %s ok.", imap->imap_user);
 
+	imap->context->cb(imap->context, DC_EVENT_IMAP_CONNECTED, 0, 0);
 	success = 1;
 
 cleanup:
--- a/src/dc_smtp.c
+++ b/src/dc_smtp.c
@@ -222,6 +222,7 @@ int dc_smtp_connect(dc_smtp_t* smtp, const dc_loginparam_t* lp)
 		dc_log_info(smtp->context, 0, "SMTP-login as %s ok.", lp->send_user);
 	}
 
+	smtp->context->cb(smtp->context, DC_EVENT_SMTP_CONNECTED, 0, 0);
 	success = 1;
 
 cleanup:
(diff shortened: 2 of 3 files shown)


src/dc_imap.c
```diff
@@ -1155,6 +1155,7 @@ static int setup_handle_if_needed(dc_imap_t* imap)
 
 	dc_log_info(imap->context, 0, "IMAP-login as %s ok.", imap->imap_user);
 
+	imap->context->cb(imap->context, DC_EVENT_IMAP_CONNECTED, 0, 0);
```
hpk42: could you pass in data2 and get rid of the dc_log_info call? 
I regard the DC_EVENT_IMAP_CONNECTED and several others events as "INFO" events as well, just specifically typed ones.  Are all events shown in the android debug log, btw?

r10s: currently, only only DC_EVENT_INFO, DC_EVENT_WARNING and DC_EVENT_ERROR events are logged automatically in android, and i think the same is true for node (cc @ralphtheninja).

the same is true for the internal logging of the last ~1000 lines that are added to the output of dc_get_info() (very handy currently as many users have no idea how to access logcat output) (i also could not say this in a few words :).

so when i would remove the dc_log_info() call, the line would not get logged.

my idea about the DC_EVENT_INFO, DC_EVENT_WARNING and DC_EVENT_ERROR events was more a sort of a wrapper around printing to stdout, stderr and so on - which are eg. Log.i(), Log.w() and Log.e() on android.

the other events in this model are mostly just state-changes that have noting to do with these log lines.

i agree that it might be smart to treat these two types the same way; if i get i correctly, you would suggest a function as `dc_log_info(context, event, data1, data2, log-string)`. maybe also a event object would be better. however, this would require some rewrite, eg. the other events do not have a string that could  be printed, some events already use data1 and data2 for other purposes, so there is no space to add another string.
also this would affect the core as well as android/android-ii and probably node - including the callback stuff.

if there are no showstoppers in the python bindings, i would suggest to postpone this, currently, there are so many open ends, i would not like to have another one :)
of course, if it helps, i could set data2 of DC_EVENT_IMAP_CONNECTED to a string what was happening.

r10s:
closed in favour of #272

PR #272, opened 2018-09-08
make send/receive message test work with python 

hpk42:
**python:** 

- add facility for running against real-life imap/smtp accounts. You need to specify a file where each line contains config settings, e.g. "addr=test1@hq5.merlinux.eu mail_pw=...." would specify one test account. For the send/receive test to work you need to have at least two lines. 

- allow to create contact and chats in the bindings 

- make travis work, running tests against py27 and py35 (but without the live-server configuration tests)

**C:** 
- c: add 3 new more specific info events so that we don't need to regex-match data2 in DC_EVENT_INFO events. 
- add a new dc_log_event function which requires specifying the DC_EVENT_* number instead of generically using DC_EVENT_INFO.  this allows to easily convert the generic into specific events. 

- this remains a bit work in progress -- if you have any tips wrt to the noted down XXX in the source code please comment there. thanks!


src/deltachat.h
```diff
@@ -1031,7 +1063,7 @@ time_t          dc_lot_get_timestamp     (const dc_lot_t*);
  */
 
 #define DC_EVENT_DATA1_IS_STRING(e)  ((e)==DC_EVENT_HTTP_GET || (e)==DC_EVENT_IMEX_FILE_WRITTEN || (e)==DC_EVENT_FILE_COPIED)
-#define DC_EVENT_DATA2_IS_STRING(e)  ((e)==DC_EVENT_INFO || (e)==DC_EVENT_WARNING || (e)==DC_EVENT_ERROR)
+#define DC_EVENT_DATA2_IS_STRING(e)  ((e)==DC_EVENT_INFO || (e) == DC_EVENT_WARNING || (e) == DC_EVENT_ERROR || (e) == DC_EVENT_SMTP_CONNECTED || (e) == DC_EVENT_IMAP_CONNECTED)
```
r10s: DC_EVENT_SMTP_MESSAGE_SENT is missing. however, maybe we can also use `(e)<1000` if we add all new "INFO" events to the area.

src/dc_chat.c
```diff
@@ -2181,6 +2181,7 @@ uint32_t dc_send_text_msg(dc_context_t* context, uint32_t chat_id, const char* t
 	uint32_t  ret = 0;
 
 	if (context==NULL || context->magic!=DC_CONTEXT_MAGIC || chat_id<=DC_CHAT_ID_LAST_SPECIAL || text_to_send==NULL) {
+        dc_log_info(context, 0, "some error");
```
r10s: nitpicking: the c-code uses tabs for indentation, after indentation, however, alignments are done with spaces.

python/tests/test_account.py
```diff
@@ -0,0 +1,72 @@
+from __future__ import print_function
+from deltachat.capi import lib
+
+
+class TestOfflineAccount:
+    def test_selfcontact(self, acfactory):
+        ac1 = acfactory.get_offline_account()
+        me = ac1.get_self_contact()
+        assert me.display_name
+        # assert me.addr  # xxx why is this empty?
```
r10s: looking in the source code, this happens if the context is not yet configured.
configure() changes and normalizes the config-options; this may also be true for the email address (spaces, maybe umlaut encoding (although not yet done) and such stuff).

python/src/deltachat/account.py
```diff
@@ -1,27 +1,157 @@
...
+            print("[{}-{}] {}({!r},{!r})".format(
+                 tname, self._logid, evt_name, data1, data2))
+
+
+class Contact:
+    def __init__(self, dc_context, contact_id):
+        self.dc_context = dc_context
+        self.id = contact_id
+        # XXX do we need to free dc_contact_t? (we own it according to API)
+        self.dc_contact_t = capi.lib.dc_get_contact(self.dc_context, contact_id)
```
r10s: yes, dc_contact_t has to be free()'d

python/src/deltachat/account.py
```diff
@@ -1,27 +1,157 @@
...
+            db_path = db_path.encode("utf8")
         capi.lib.dc_open(ctx, db_path, capi.ffi.NULL)
-        self._logcallback = logcallback
+        self._evhandler = EventHandler(self.dc_context)
+        self._evlogger = EventLogger(self.dc_context, _logid)
+        self._threads = IOThreads(self.dc_context)
 
     def set_config(self, **kwargs):
         for name, value in kwargs.items():
+            name = name.encode("utf8")
```
flub: this really smells like python2 unicode troubles going to hurt badly at some point...  i'll repeat my "please very carefully consider if python 2 support won't do more harm then good" plea...

python/src/deltachat/account.py
```diff
@@ -1,27 +1,157 @@
...
 
+    def get_config(self, name):
+        name = name.encode("utf8")
+        res = capi.lib.dc_get_config(self.dc_context, name, b'')
+        return ffi_unicode(res)
+
+    def get_self_contact(self):
+        return Contact(self.dc_context, capi.lib.DC_CONTACT_ID_SELF)
+
+    def create_contact(self, email, name=ffi.NULL):
```
flub: since this is aiming at being the higher level layer the API interface should hide all cffi bits IMHO, so a default of `name=None` is better in i think

python/src/deltachat/account.py
```diff
@@ -31,38 +161,17 @@ def shutdown(self):
         # interrupt idle calls do not seem to release the
         # blocking call to smtp|imap idle. This means we
         # also can't now close the database because the
-        # threads might still need it
+        # threads might still need it.
         # capi.lib.dc_close(self.dc_context)
 
-    def process_event(self, ctx, evt_name, data1, data2):
+    def _process_event(self, ctx, evt_name, data1, data2):
```
flub: When reading this function I don't actually know what's a cffi/c object and what's a python object.  I think it's important to make the distinction very clearly and have a clear boundary so that I don't have to wonder about this.  E.g. is `_evtlogger` taking cffi objects or python ones?  I feel like it should be python ones since converting should happen as soon as it's possible to do so accurately.

python/src/deltachat/account.py
```diff
@@ -104,3 +213,15 @@ def smtp_thread_run(self):
         while not self._thread_quitflag:
             capi.lib.dc_perform_smtp_jobs(self.dc_context)
             capi.lib.dc_perform_smtp_idle(self.dc_context)
+
+
+def convert_bytes(obj):
```
flub: convert what to what?  I mean naming and signature matters, docstrings help.  Same for ffi_unicode.  But also these two are clearly cffi implementation bits, the user shouldn't see this, why are these functions public?

src/dc_chat.c
flub: +1 for nitpicking and +1 for consistency

-8 for having chosen tabs :stuck_out_tongue:

flub:
I forgot to add, but this does a lot of custom ad-hoc logging which reminds me of the various "quick" logging hacks which haunted pytest for so long.  I think it'd be much better to adopt something more standardised and thought out.  I think structlog (http://www.structlog.org) might be a good candidate by the (very very quick) look of what you're trying to do with passing logids around.

python/src/deltachat/account.py
hpk42: in python 3 kwargs always come in as unicode, and we need to encode this for the delta c-api which does not know about unicode.  This function is about config options -- we could probably just encode with "ascii" to be sure nobody tries non-ascii config names.

python/src/deltachat/account.py
```diff
@@ -1,27 +1,157 @@
...
+
 from . import capi
-import deltachat
 from .capi import ffi
+import deltachat
+
+
+class EventHandler:
+    def __init__(self, dc_context):
+        self.dc_context = dc_context
```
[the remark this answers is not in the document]
hpk42: for now, i do it in general -- if a subclass wants to do something with the capi it needs it.

python/src/deltachat/account.py
hpk42: otherwise i agree that moving to py3 makes sense, just not there yet becaus ei am dependent on an old project (rlcompleter2) which i wrote around 2000 and which i still heavily use ... and it does not work on python3 ...

hpk42: that being said, i think there are still quite some people i guess who have older systems and are happy if py2 bindings work.  i'd like to use the bindings server-side btw.

python/src/deltachat/account.py
hpk42: yes, these are python objects here.  the py_dc_callback converts them early on. so all occurences of evt_name, data1, data2 etc. in account.py are python objects.

python/src/deltachat/account.py
```diff
@@ -1,27 +1,157 @@
...
+        self._event_queue = Queue()
+        self._debug = debug
+        if _logid is None:
+            _logid = str(self.dc_context).strip(">").split()[-1]
+        self._logid = _logid
+        self._timeout = None
+
+    def __call__(self, evt_name, data1, data2):
+        self._log_event(evt_name, data1, data2)
+        self._event_queue.put((evt_name, data1, data2))
```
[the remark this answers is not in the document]
hpk42: they are python objects.

hpk42:
added an issue about structlog and attr https://github.com/deltachat/deltachat-core/issues/273
don't want to make this part of this PR.

hpk42:
also added https://github.com/deltachat/deltachat-core/issues/274 for docstrings that specify which objects are expected.

python/tox.ini
```diff
@@ -0,0 +1,39 @@
...
+    # pygments required by rst-lint
+    pygments
+    restructuredtext_lint
+commands =
+    flake8 src/deltachat
+    flake8 tests/
+    # {envpython} scripts/check-rst.py
+
+[pytest]
+minversion = 2.0
```
flub: do you still run with 2.x?  I mean, is this actually tested?  pytest has been 3.x for a while now...

Issue #276, opened 2018-09-09
Clean up installed headers


flub:
Currently we install the main public header `deltachat/deltachat.h`, but we also install the various headers which I thought where private in the same directory, e.g. `deltachat/dc_smtp.h`.  Once installed someone who tries to use one of the `dc_*.h` headers can not actually do so since they include each other directly like `#include "dc_smtp.h"` which does not work on the installed headers (where you'd have to do `#include <deltachat/dc_smtp.h>`.

I think the most logical thing would be to simply not install these private headers anymore.  The other option is to make them public and make them all import each other with `#include <deltachat/dc_*.h>`.

Which one shall we do?

r10s:
the dc_*.h files are private to the deltachat-core library, i think they should not be used from extern, so i think there is no need to install them (if there is a need, this seems to be a bug :).

PR #277, opened 2018-09-10
Do not install internal header files

flub:
Fixes issue #276


ralphtheninja:
What does this really prevent? Installing these headers when `ninja install` is called?

flub:
Yes

PR #278, opened 2018-09-12
Add option to add a special marker for "All done" to the chatlist

r10s:
this is a helper for the UI creating the chatlist.

already today, the chatlist consist out of normal chats identified by an ID _plus_ some special IDs representing special items. currently, these items are the "Archive" link and the "Contact requests".

UIs as the new signal port show a help-hint if everything is archived (sth. as _"well done, nothing more to do ... you have a zero inbox"_. to help these UIs adding such an item to the virtual "recycler" view - which is typically complicated enough - this pr offers an option to easily add such an item as needed:
 
if the flag `DC_GCL_ADD_ALLDONE_HINT` is given to `dc_get_chatlist()` and there are no but archived chats, the item `DC_CHAT_ID_ALLDONE` is added; typically together with the "Archive" link.

### DIFF
--- a/src/dc_chatlist.c
+++ b/src/dc_chatlist.c
@@ -379,6 +379,10 @@ static int dc_chatlist_load_from_db(dc_chatlist_t* chatlist, int listflags, cons
 
     if (add_archived_link_item && dc_get_archived_cnt(chatlist->context)>0)
     {
+		if (dc_array_get_cnt(chatlist->chatNlastmsg_ids)==0 && (listflags & DC_GCL_ADD_ALLDONE_HINT)) {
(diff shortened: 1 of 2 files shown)


r10s:
thanks for merging :)

testbird:
> UIs as the new signal port show a help-hint if everything is archived (sth. as "well done, nothing more to do ... you have a zero inbox" [...] typically together with the "Archive" link

Now that makes more sense, but strange why such changing state is called "archived" and not simply "done" or "inactive".

r10s:
messenger-land is quite different form e-mail-land here and there :)

Issue #273, opened 2018-09-09
introduce attr and probably structlog


hpk42:
As pointed out by @flub consider introducing structlog (and also attr) to get some more standard handling of object properties/attributes and logging.

hpk42:
while attr is in use now, i remain skeptical of trying to use structlog here.  My tries so far (a few hours) did not really leave me satisfied but i haven't grasped structlog fully yet.  What i am interested in: 

- logging should in a human-readable manner print thread and dc_context contexts 
- it should be nicely useable to test arrival of logging events (i.e. used from test suite)
- have a non-timestamp formatting, at last optionally  

it's quite possible i am missing something, though.  So i changed the title of this issue to "play further with structlog" ...

PR #279, opened 2018-09-13
 introduce attr / refine GC handling

hpk42:
address #273 -- also refine the GC-handling of contact and message pointers.
free-ing dc_context pointers kind of works now -- but without calling account.shutdown() first 
then GC ``__del__`` calls could end up without access to globals. Refining/reviewing GC handling should mostly be part of a separate effort than this PR.  The current PR is at least an improvement.

### DIFF
--- a/python/src/deltachat/account.py
+++ b/python/src/deltachat/account.py
@@ -112,7 +112,7 @@ class Chat(object):
 
     @cached_property
     def dc_chat_t(self):
-        return capi.lib.get_chat(self.id)
+        return capi.lib.dc_get_chat(self.dc_context, self.id)
 
     def __del__(self):
         capi.lib.dc_chat_unref(self.dc_chat_t)
@@ -131,7 +131,7 @@ class Message(object):
     id = attrib_int()
 
     @cached_property
-    def dc_msg(self):
+    def dc_msg_t(self):
         return capi.lib.dc_get_msg(self.dc_context, self.id)
 
     def __del__(self):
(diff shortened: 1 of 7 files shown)


python/src/deltachat/myattr.py
```diff
@@ -0,0 +1,33 @@
+from attr import attrs, attrib  # noqa
+from attr import validators as v
+
+
+def attrib_int():
+    return attrib(validator=v.instance_of(int))
+
+
+def attrib_CData():
+    from deltachat.capi import ffi
```
flub: why does this need to be in the function?

python/src/deltachat/myattr.py
```diff
@@ -0,0 +1,33 @@
...
+    return attrib(validator=v.instance_of(ffi.CData))
+
+
+# copied over unmodified from
+# https://github.com/devpi/devpi/blob/master/common/devpi_common/types.py
+# where it's also tested
```
flub: hah, should also copy the test clearly!

python/src/deltachat/account.py
```diff
@@ -7,14 +7,15 @@
 except ImportError:
     from Queue import Queue
 
+import deltachat
 from . import capi
 from .capi import ffi
-import deltachat
+from .myattr import attrib_CData, attrs, attrib_int, cached_property
```
flub: I'm not a huge fan of the attrib_* shortcuts, they're not very verbose to write and the shortcuts hide things somewhat.  Likewise importing attris/attrib via myattr is less clear then using `attr.s` `attr.ib` directly, now I have to worry about whether they're actually what I think they'll be which wouldn't be the case if importing `attr` directly.

I'm generally not a fan of importing objects directly btw, always prefer importing modules and accessing objects inside them via the module.  But clearly you don't share that view so I don't mind, just thought I'd mention it.

Anyway, all this is a style decision really, I've argued the readability point of view but up to you in the end.

python/src/deltachat/account.py
```diff
@@ -151,6 +159,9 @@ def __init__(self, db_path, logid=None):
         self._evlogger = EventLogger(self.dc_context, logid)
         self._threads = IOThreads(self.dc_context)
 
+    def __del__(self, dc_context_unref=capi.lib.dc_context_unref):
```
flub: Is it not better to just not do anything if the module is `None`?  the interpreter is already shutting down anyway... (sure this breaks some sub-interpreter use-cases, but hey)

also i believe this problem magically goes away when using Python 3.x+ (for x being something like 4, 5 or 6) ;)

PR #281, opened 2018-09-14
next round of python bindings update

hpk42:
- initial set of sphinx docs 
- docstrings on most functions and the account module (might be split sometimesoon) 
- include doc generation into tox run 
- fix attrs dependency  to work with/without pytest
- don't skip offline-account tests unncessarily
- implement Chat.get_messages(), internally dealing with dc_array_t


python/src/deltachat/account.py
```diff
@@ -266,5 +410,13 @@ def convert_to_bytes_utf8(obj):
     return obj
 
 
+def iter_array_and_unref(dc_array_t, constructor):
+    try:
+        for i in range(0, lib.dc_array_get_cnt(dc_array_t)):
+            yield constructor(lib.dc_array_get_id(dc_array_t, i))
```
flub: Is the underlying deltachat data here actually a fully populated array with all data or an array of IDs/pointers or an iterator?  I ask because if it's not the first it might be worth creating an DCArrayIterator object or so which is actually a real iterator (implementing collections.abc.Iterator), or maybe a DCArrayContainer if it's indexable or can be iterated multiple times (implementing collections.abc.Collection or whatever is appropriate)

r10s: it's an fully populated array, so `dc_array_get_id()` is cheap, only a function call and a return a[i], no i/o is involved here.
typical arrays have some 1000 entries (eg. a longer chat history) most are much smaller, typically not one million or so :)

hpk42:
On Fri, Sep 14, 2018 at 04:42 -0700, björn petersen wrote:
> r10s commented on this pull request.
> 
> 
> 
> > @@ -266,5 +410,13 @@ def convert_to_bytes_utf8(obj):
>      return obj
>  
>  
> +def iter_array_and_unref(dc_array_t, constructor):
> +    try:
> +        for i in range(0, lib.dc_array_get_cnt(dc_array_t)):
> +            yield constructor(lib.dc_array_get_id(dc_array_t, i))
> 
> it's an fully populated array, so `dc_array_get_id()` is cheap, only a function call and a return a[i], no i/o is involved here.
> typical arrays have some 1000 entries (eg. a longer chat history) most are much smaller, typically not one million or so :)

if it ever proves to be a performance problem, we can also
introduce additional query-parameters like date-range or "last 50" or whatever. 
Or is there an API for that already? haven't fully explored the c-api yet ... 
Oh, and if we use pypy the py/c boundary performance hit would go away ;)

PR #282, opened 2018-09-14
improve docs and structure

hpk42:
see online snapshot: https://m.devpi.net/hpk/dev/deltachat/0.5.dev0/+doc/index.html

### DIFF
--- a/python/.gitignore
+++ b/python/.gitignore
@@ -1,0 +1,1 @@
+doc/_build/
--- a/python/doc/_templates/globaltoc.html
+++ b/python/doc/_templates/globaltoc.html
@@ -4,7 +4,8 @@
 <ul>
   <li><a href="{{ pathto('index') }}">Home</a></li>
   <li><a href="{{ pathto('install') }}">Install</a></li>
-  <li><a href="{{ pathto('api') }}">API Documentation</a></li>
+  <li><a href="{{ pathto('api') }}">High level API</a></li>
+  <li><a href="{{ pathto('capi') }}">Low level API</a></li>
 </ul>
 <b>External Links:</b>
 <ul>
--- a/python/doc/_templates/globaltoc.html
+++ b/python/doc/_templates/globaltoc.html
@@ -1,17 +1,15 @@
 
-<h3><a href="{{ pathto('contents') }}">{{ _('Table Of Contents') }}</a></h3>
-
 <ul>
-  <li><a href="{{ pathto('index') }}">Home</a></li>
-  <li><a href="{{ pathto('install') }}">Install</a></li>
-  <li><a href="{{ pathto('api') }}">High level API</a></li>
-  <li><a href="{{ pathto('capi') }}">Low level API</a></li>
+  <li><a href="{{ pathto('index') }}">index</a></li>
+  <li><a href="{{ pathto('install') }}">install</a></li>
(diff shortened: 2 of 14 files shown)


hpk42:
would be nice if https://m.devpi.net/hpk/dev/deltachat/0.5.dev0/+doc/capi.html showed the docstrings of the c-api ... @flub any idea if we could somehow parse/preserve the docstrings in some (i guess hackish) way?  I'd like this link to become https://py.delta.chat btw ...

flub:
Not sure how you'd do that. Maybe doxygen (I believe that's the current C
format?) Can export to something like XML or so that can be transformed? Is
there a python API for doxygen or do you need to script things?

hpk42:
On Sat, Sep 15, 2018 at 13:07 -0700, Floris Bruynooghe wrote:
> Not sure how you'd do that. Maybe doxygen (I believe that's the current C
> format?) Can export to something like XML or so that can be transformed? Is
> there a python API for doxygen or do you need to script things?


This page: https://www.stack.nl/~dimitri/doxygen/manual/output.html
seems to suggest it's possible to generate xml.  

Supposing we can parse the C-docstrings this way, how could we 
make them available with the low-level capi.lib functions
(and then through sphinx)?  Could this be done during _build.py time? 
doing this at import-time is obviously not nice.

flub:
Hum, I think that'd be a feature request for cffi as nothing in the docs
seems present.  In fact even import time is not going to work since the
__doc__ already contains the C signature and appears immutable.

Is this really worth the effort though?  Users shouldn't be using this
and people working on deltachat itself can refer to the C docs anyway.

hpk42:
On Sun, Sep 16, 2018 at 10:16 -0700, Floris Bruynooghe wrote:
> Hum, I think that'd be a feature request for cffi as nothing in the docs
> seems present.  In fact even import time is not going to work since the
> __doc__ already contains the C signature and appears immutable.

there still is a lot of delta-core C functions to develop against ... 
i kind of like browsing https://m.devpi.net/hpk/dev/deltachat/0.5.dev2/+doc/capi.html?highlight=dc_open# -- staying on the page thre with search etc ... 
but didn't know it requires changing CFFI ... 

> Is this really worth the effort though?  Users shouldn't be using this
> and people working on deltachat itself can refer to the C docs anyway.

yes, it's workable.

flub:
On Sun 16 Sep 2018 at 11:46 -0700, holger krekel wrote:

> On Sun, Sep 16, 2018 at 10:16 -0700, Floris Bruynooghe wrote:
>> Hum, I think that'd be a feature request for cffi as nothing in the docs
>> seems present.  In fact even import time is not going to work since the
>> __doc__ already contains the C signature and appears immutable.
>
> there still is a lot of delta-core C functions to develop against ... 
> i kind of like browsing https://m.devpi.net/hpk/dev/deltachat/0.5.dev2/+doc/capi.html?highlight=dc_open# -- staying on the page thre with search etc ... 
> but didn't know it requires changing CFFI ... 

Depends what the actual goal is here.  If you don't care about the
docstrings in the module but care about the docs showing up in sphinx
then maybe you shouldn't give up yet.  Maybe you can do something in
conf.py or some slightly less hacky place in sphinx where you run
docxygen to produce xml-output (or whatever most relevant) and then
parse the xml to output an c-api.rst file which is then included in the
sphinx toctree.

PR #285, opened 2018-09-15
lookup contacts

r10s:
added two functions:
- `dc_lookup_contact_id_by_addr()` checks if an e-mail-address is in the database (wondering that this function does not exist :)
- `dc_may_be_valid_addr()` is used internally to check addresses eg. given to dc_create_contact(). thought that i would need it for the new android port, however, it's not needed. nevertheless, it may be useful here and there - esp. as it is the same logic as used internally to validate addresses.

nb: the validation is rather generous, however, doing more complex regex things will increase problems but probably not solve some. validating email addresses "the right way" seems to be very hard, see eg. https://www.regular-expressions.info/email.html

### DIFF
--- a/cmdline/stress.c
+++ b/cmdline/stress.c
@@ -359,6 +359,20 @@ void stress_functions(dc_context_t* context)
 	 **************************************************************************/
 
 	{
+		assert( !dc_may_be_valid_addr(NULL) );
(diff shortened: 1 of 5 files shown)


r10s:
meging myself as i do not want to leave the android port in a state that cannot be compiled (it needs `dc_lookup_contact_id_by_addr()`)

PR #294, opened 2018-09-19
add some more high-level API, and tests

hpk42:
- refactor gc-handling wrt to array
- don't cache dc_chat and dc_msg and dc_contact structs -- "if in doubt, don't cache" as the wise programmers say. and i am not sure we need it.  the caching was also there to handle GC things but now that we have "ffi.gc()" we don't need that 
- add some high level APIs (forwarding messages, setting chat names, creating group chats and getting all chats ... but the latter are very simple chats for now, do not expose the full underlying API)
- improve some doc strings

### DIFF
--- a/python/src/deltachat/chatting.py
+++ b/python/src/deltachat/chatting.py
@@ -2,7 +2,7 @@
 
 from .cutil import as_dc_charpointer, from_dc_charpointer, iter_array
 from .capi import lib, ffi
-from .types import cached_property, property_with_doc
+from .types import property_with_doc
 import attr
 from attr import validators as v
 
@@ -16,7 +16,7 @@ class Contact(object):
     _dc_context = attr.ib(validator=v.instance_of(ffi.CData))
     id = attr.ib(validator=v.instance_of(int))
 
-    @cached_property
+    @property
     def _dc_contact(self):
         return ffi.gc(
             lib.dc_get_contact(self._dc_context, self.id),
@@ -51,7 +51,7 @@ class Chat(object):
     _dc_context = attr.ib(validator=v.instance_of(ffi.CData))
     id = attr.ib(validator=v.instance_of(int))
 
-    @cached_property
+    @property
     def _dc_chat(self):
         return ffi.gc(
             lib.dc_get_chat(self._dc_context, self.id),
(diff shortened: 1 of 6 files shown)


python/src/deltachat/chatting.py
```diff
@@ -94,6 +129,34 @@ def mark_noticed(self):
...
+
+    def get_contacts(self):
+        """ get all contacts for this chat.
+
+        :params: contact object.
+        :exception: ValueError if contact could not be added
+        :returns: None
+        """
+        dc_array = ffi.gc(
+            lib.dc_get_contacts(self._dc_context, lib.DC_GCL_ADD_SELF, ffi.NULL),
```
r10s: `dc_get_contacts()` returns all available contacts.
to get the contacts for a given chat, use dc_get_chat_contacts() - which takes the chat id as the second paramter then

python/tests/test_account.py
```diff
@@ -51,9 +51,30 @@ def test_chat(self, acfactory):
...
+
+    def test_group_chat_creation(self, acfactory):
+        ac1 = acfactory.get_offline_account()
+        contact1 = ac1.create_contact("some1@hello.com", name="some1")
+        contact2 = ac1.create_contact("some2@hello.com", name="some2")
+        chat = ac1.create_group_chat(name="title1")
+        chat.add_contact(contact1)
+        chat.add_contact(contact2)
+        assert chat.get_name() == "title1"
+        assert contact1 in chat.get_contacts()
```
r10s: not sure was the in really does and no concrete idea about how python objects work, it may just be that we're comparing different objects here where we want to compare the ids.

eg. would `contact1 = ac1.create_contact("same@same.same")` and `contact2 = ac1.create_contact("same@same.same")` return the _same_ objects? the ids returned by `dc_create_contact()` will be the same.

so `contact1.id == contact2.id` will work, not sure if `contact1 == contact2` is true or false.

python/src/deltachat/chatting.py
hpk42: ok.  but wouldn't dc_get_chat_contacts() always return a subset of or the same as dc_get_contacts()?

python/tests/test_account.py
hpk42: don't worry -- "contact1 == contact2" implies "contact1.id == contact2.id and contact1._dc_context == contact2._dc_context.

PR #295, opened 2018-09-19
explicitly handle empty strings received from the UI on DC_EVENT_HTTP_GET

r10s:
up to now, "Cannot read file" errors are only printed when DC_EVENT_HTTP_GET returns NULL.
if an empty string was returned, deltachat-core tries to parse this as xml content - which, of course, also results in an error some time later.

from the view of deltachat-core, there is no need to distinguish between NULL or an empty string.

so, this pr accepts empty strings explicitly as an error and also documents be behavior so that it is easier for bindings to return this error state.

@ralphtheninja when this is merged, i think we can remove the explicit handling of empty stirngs in [dcn_set_http_get_response](https://github.com/deltachat/deltachat-node/blob/master/src/module.c#L1331)

### DIFF
--- a/src/dc_configure.c
+++ b/src/dc_configure.c
@@ -38,12 +38,16 @@
 static char* read_autoconf_file(dc_context_t* context, const char* url)
 {
 	char* filecontent = NULL;
+
 	dc_log_info(context, 0, "Testing %s ...", url);
+
 	filecontent = (char*)context->cb(context, DC_EVENT_HTTP_GET, (uintptr_t)url, 0);
-	if (filecontent==NULL) {
+	if (filecontent==NULL || filecontent[0]==0) {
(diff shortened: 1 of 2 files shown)


src/dc_configure.c
```diff
@@ -38,12 +38,16 @@
 static char* read_autoconf_file(dc_context_t* context, const char* url)
 {
 	char* filecontent = NULL;
+
 	dc_log_info(context, 0, "Testing %s ...", url);
+
 	filecontent = (char*)context->cb(context, DC_EVENT_HTTP_GET, (uintptr_t)url, 0);
-	if (filecontent==NULL) {
+	if (filecontent==NULL || filecontent[0]==0) {
```
ralphtheninja: :+1:

PR #305, opened 2018-09-24
suspend smtp-thread by waiting-for-job-finish instead of entering-idle

### LINKED ISSUE #304: dc_configure() should work without threads being started
in a normal environment, both threads will be running when dc_configure() is called.
however, for testing it is convenient to do this without threads @hpk42 .

currently, dc_configure() does not work if there is no smtp-thread,
the reason for this is that the whole configuration runs in the imap thread and the smtp-thread is suspended  by waiting until it is in idle-mode and left there.
if the smtp-thread is not threre, dc_configure() waits infinitly as the smtp-thread will never go into the idle-mode ([flag smtpidle_in_idleing is never set](https://github.com/deltachat/deltachat-core/blob/master/src/dc_job.c#L483))

r10s:
to allow dc_configure() being called without the threads being started, we suspend the smtp-thread now by checking that we're not in performing jobs instead of waiting for entering idle.

closes #304 if being merged

### DIFF
--- a/src/dc_context.h
+++ b/src/dc_context.h
@@ -77,8 +77,8 @@ struct _dc_context
 	pthread_cond_t   smtpidle_cond;
 	pthread_mutex_t  smtpidle_condmutex;
 	int              smtpidle_condflag;
-	int              smtpidle_suspend;
-	int              smtpidle_in_idleing;
+	int              smtp_suspended;
+	int              smtp_doing_jobs;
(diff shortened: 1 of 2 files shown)


r10s:
however, this will probably also require the need to suspend the smtp-thread as at a given point we need to swap/reset/reconfigure the context's smtp object. we cannot do this when the smtp-thread is performing jobs

Issue #301, opened 2018-09-24
Remove dep to bzip2


angelo-fuchs:
Bzip is unused in deltachat, so it can be removed

PR #310, opened 2018-09-25
Update netpgp build

flub:
This does auto-configuration of the netpgp dependency and makes it build without the bzip2 dependency in delta-core.  It's my version of fixing #301 :)

I believe this is an improvement for handling the netpgp dependency and the other deps should follow this too.


libs/netpgp/meson.build
```diff
@@ -2,14 +2,70 @@ project('netpgp', 'c',
...
+
+## dependencies
 pthreads = dependency('threads')
 zlib = dependency('zlib', fallback: ['zlib', 'zlib_dep'])
 openssl = dependency('openssl', fallback: ['openssl', 'dep'])
 
-# Dependencies without pkg-config, just try linking against them
-cc = meson.get_compiler('c')
+# BZip2 has no pkg-config, just try linking against it
+bzip2 = cc.find_library('bz2', required: get_option('bzip2'))
```
hpk42: why do we still link against bzip2 here?

flub:
Bzip2 is now a feature in meson_options.txt which means you can enable,
disable or autodetect it. This only looks for the dependency if it's not
disabled and makes it optional in that auto case. If you look at how the
main meson.build uses this subproject you'll see it disables the feature,
which is what we want for delta-core

r10s:
hi @flub thank you for pushing this forward.

i am wondering how i can create an config-netpgp.h for android, and also for ios.
maybe we can include them using <config-netpgp.h> (`<..>` instead of `".."`) so that we can define a project-specific file without going through meson.

also, i've misused some of the defines in config-netpgp.h as a guarantee about what should work (eg. HAVE_OPENSSL_CAMELLIA_H - it was always just defined to 1 and if the code is not there, compilation would break).
how can i do this now? maybe by a checking the algos at runtime?

apart from these questions, i've tried a build and things do work on my ubuntu 18.04 :)

flub:
So afaik short-turm delta-android is not affected by these things in delta-core since they still use their own copy of this (same) code.  Correct me if I'm wrong and this is no longer the case though.

For the long term, O(months), I was aiming for the meson build system of detla-core to be able to be cross-compiled and producing a number of lib*.a files which the android build would reference directly.  So in that case the meson build has to produce the right config-netpgp.h file for android which it would probably do very similarly to how this is done here - probably some tweaks added.

WRT to the HAVE_OPENSSL_CAMELLIA_H, it means I need to turn the openssl-camellia and openssl-idea options into features as well so that the delta-core build can explicitly control which ones are needed and which not.  I'll update this PR with that.

libs/netpgp/meson.build
flub: huh, apparently email replies don't thread terribly well.  https://github.com/deltachat/deltachat-core/pull/310#issuecomment-424704208 was meant to be a response to this

r10s:
> So afaik short-turm delta-android is not affected by these things in delta-core

deltachat-android is using deltachat-core as a subproject, so it has the "same" files.
however, i just tried to ndk-build deltachat-android with the changes of this pr - it works by just adding the old netpgp-config.h to the jni subdirectory :)

so, from the android-side, this pr is fine if CAMELLIA is enforced. ios is not updated that frequently currently, but it will probably work the same way.

flub:
Oh, I didn't realise you actually switched deta-android to use the code from here already.  I'll have to be a little more careful then to not break the existing code.

I'll put the original config-netpgp.h back in place and figure something else out for the meson build.  Maybe just order of include directories is enough but that could be too brittle.

flub:
I think I've addressed the comments now.  Thanks for looking carefully!

PR #311, opened 2018-09-26
Hack around libetpan not respecting wrap-mode

flub:
Since we do not have access to the value of wrap-mode we can not
make the etpan dependency to behave correctly.  Hack around it by
introducing a new option just for etpan.

@ralphtheninja this would need your node bindings to use the new -Dforce-etpan-fallback=true as well as --wrap-mode=forcefallback.  But would mean it should correctly ignore the system one.  What do you think?

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -33,12 +33,14 @@ endif
 
 # Sadly libetpan does not use pkg-config.  Use the system one if it's
 # new enough, if not use a static-pic-lib regardless of
-# default-library and wrap-mode.
+# default-library.
 libetpan_config = find_program('libetpan-config', required: false)
 if (not get_option('monolith')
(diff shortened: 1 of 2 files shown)


ralphtheninja:
@flub What if I no longer want to use `--wrap-mode=forcefallback` (want to build without bundled openssl) can I still use `-Dforce-etpan-fallback=true` to get the bundled libetpan?

flub:
Yes, you can. Though libetpan is special now since it's the only dependency
which can be toggled like this. Which makes me wonder if this is a good
idea after all.

r10s:
lgtm

flub:
@ralphtheninja I'll leave merging to you since you're most affected

PR #316, opened 2018-09-26
tox already cares for doc-building

hpk42:
simplify doc building, cc @azul

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,11 +41,6 @@ script:
   - pip install tox
   - ldd /usr/local/lib/x86_64-linux-gnu/libdeltachat.so
   - tox
-  - pip install -e .
-  - pip install sphinx breathe
-  - cd doc
-  - make html
-  - cd ../..
 
 deploy:
   provider: script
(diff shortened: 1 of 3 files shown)


azul:
uhh... nice.

.travis.yml
```diff
@@ -41,11 +41,6 @@ script:
   - pip install tox
   - ldd /usr/local/lib/x86_64-linux-gnu/libdeltachat.so
   - tox
-  - pip install -e .
-  - pip install sphinx breathe
-  - cd doc
-  - make html
-  - cd ../..
```
azul: you will still need to `cd ..`.
Otherwise the script ends in a subdir and the deploy step tries to deploy from there.
I know it's stupid.
Maybe `cd $TRAVIS_BUILD_DIR` at the top of teh deploy step is the better solution.

azul: found and applied a more robust solution.

azul:
@hpk42 I updated this so the script can be called.
Also removed the unneccessary gem.

From my point of view this is ready to go.

hpk42:
thanks a lot, @azul !

PR #321, opened 2018-09-27
Try doxygen new

azul:
Fixed the ppa issue.

However this still only gives us doxygen 1.8.7, while @hpk42 said we need 1.8.11.

Looks like there is no ppa for 1.8.11: https://www.ubuntuupdates.org/pm/doxygen


r10s:
i think due to #325 this is no longer needed can can be closed without merging

r10s:
:)

PR #332, opened 2018-09-28
Version the libdeltachat.so file

flub:
This creates a version scheme extracted from the existing git tagging
schema being used.  It therefore relies on this schema not changing.

Since there appear no way to do all of this reading things from git
before the project() call we can sadly not use the correct version
in the project itself.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -12,6 +12,7 @@ project(
 #  - X.Y.Z for tagged releases.
 #  - X.Y.Z990N for dev releases.
 #    Where N is the number of commits since the last tag.
+version = meson.project_version()
 git = find_program('git', required: false)
 if git.found()
   git_desc = run_command(git, 'describe', '--tags')
@@ -24,14 +25,10 @@ if git.found()
                           version_parts[1],
                           version_parts[2] + '990' + git_desc_parts[1]])
     endif
-  else
-    version = meson.project_version()
   endif
-else
-  version = meson.project_version()
 endif
 if version == meson.project_version()
-  warning('Git version not foud, using (dummy) project version')
+  warning('Git version not found, using (dummy) project version')
 endif
 
 
--- a/src/meson.build
+++ b/src/meson.build
@@ -65,4 +65,4 @@ pkg.generate(libraries : lib,
              subdirs : ['deltachat'],
              name : 'libdeltachat',
              filebase : 'deltachat',
-             description : ' Create your own, email-compatible messenger.')
+             description : 'Create your own, email-compatible messenger.')


meson.build
```diff
@@ -1,10 +1,40 @@
...
+                          version_parts[2] + '990' + git_desc_parts[1]])
+    endif
+  else
+    version = meson.project_version()
+  endif
+else
+  version = meson.project_version()
+endif
+if version == meson.project_version()
+  warning('Git version not foud, using (dummy) project version')
```
ralphtheninja: `foud` -> `found`

src/meson.build
```diff
@@ -46,12 +46,23 @@ lib = library(
...
                 subdir: 'deltachat')
+
+
+pkg = import('pkgconfig')
+pkg.generate(libraries : lib,
+             version : version,
+             subdirs : ['deltachat'],
+             name : 'libdeltachat',
+             filebase : 'deltachat',
+             description : ' Create your own, email-compatible messenger.')
```
ralphtheninja: Space before `Create`

meson.build
```diff
@@ -1,10 +1,40 @@
...
+      version_parts = version.split('.')
+      version = '.'.join([version_parts[0],
+                          version_parts[1],
+                          version_parts[2] + '990' + git_desc_parts[1]])
+    endif
+  else
+    version = meson.project_version()
+  endif
+else
+  version = meson.project_version()
```
ralphtheninja: Maybe set `version = meson.project_version()` before `if git.found()` and skip all the else cases?

flub: why didn't i think of that?  thanks!

muelli:
is this about naming the .so file on disk? In that case, the ABI should probably play a role in determining the version.

flub:
Yes it is.  And this does rely on the tagging to happen correctly, the commit message mentions that the major version number **must** change when the ABI changes in an incompatible way.  But it's a good point, is the main tag version happy to be used like this?  If you do this strictly you can quickly end up with high version numbers.

flub:
So @r10s: any opinions on repo tagging vs ABI versions.  Do you think you can keep this in sync?  If not do you think you can remember to manually bump the ABI every time it changes independently of the tagged repo version?

Technically you could argue you only need to bump the ABI/soname version at the time of a release (which I take to be the time when you tag the repo).  But then at this time you need to decide if the ABI changed since the last tagged release or not and bump the ABI/soname version independently.  This is certainly more of a pain.

Alternatively the cheap way out is to say that the ABI has no guarantees at all while it's not yet 1.0.  Though it's only postponing this problem so maybe best to figure it out now.

flub:
björn petersen <notifications@github.com> writes:

>> Alternatively the cheap way out is to say that the ABI has no guarantees at all while it's not yet 1.0. 
>
> do you mean just to drop dc_get_version() and the corresponding
> DC_VERSION_STR define?

No, I didn't mean that.  The way that's used is perfectly fine I think,
it's just the soversion which we may not be treating correctly.  But I'm
fine with ignoring it until someone starts shouting at us for not
behaving correctly ;)

r10s:
okay :)

Issue #333, opened 2018-09-29
app crashes Invalid address 0x93968e00 passed to free: value not allocated libmessenger.1.so (CRYPTO_free+24)


muelli:
**Delta Chat version**

dunno. I can't access the app, because it crashes. But f-droid doesn't show updates.

**Expected behavior**

I expected to be able to browse my messages.

**Actual behavior**

app crashes.

**Steps to reproduce the problem**

1. open app

**Screen snapshots**

**Debug logs**
<details>
<summary>Logs</summary>

```

<!--
If applicable, debug logs can be copied from within the Delta Chat app: 

Open the _Settings_ menu -> _About_ -> _version number|Info_.

Alternatively from the Android system log:

` adb logcat -v time -s DeltaChat `

This log contains private data (e.g. mail address, provider information) which shall be removed or anonymised prior to posting.
-->

```
</details>

```
09-29 11:38:59.572 23521 23538 F libc    : Invalid address 0x93968e00 passed to free: value not allocated
09-29 11:38:59.572 23521 23538 F libc    : Fatal signal 6 (SIGABRT), code -6 in tid 23538 (smtpThread)
09-29 11:38:59.573   382   382 W         : debuggerd: handling request: pid=23521 uid=10095 gid=10095 tid=23538
09-29 11:38:59.639 23566 23566 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
09-29 11:38:59.639 23566 23566 F DEBUG   : Build fingerprint: 'bq/Aquaris_X5_Plus/Aquaris_X5_Plus:7.1.1/NMF26F/676:user/release-keys'
09-29 11:38:59.639 23566 23566 F DEBUG   : Revision: '0'
09-29 11:38:59.639 23566 23566 F DEBUG   : ABI: 'arm'
09-29 11:38:59.639 23566 23566 F DEBUG   : pid: 23521, tid: 23538, name: smtpThread  >>> com.b44t.messenger <<<
09-29 11:38:59.639 23566 23566 F DEBUG   : signal 6 (SIGABRT), code -6 (SI_TKILL), fault addr --------
09-29 11:38:59.643 23566 23566 F DEBUG   : Abort message: 'Invalid address 0x93968e00 passed to free: value not allocated'
09-29 11:38:59.643 23566 23566 F DEBUG   :     r0 00000000  r1 00005bf2  r2 00000006  r3 00000008
[...]

PR #334, opened 2018-09-29
check for some NULL-pointers in netpgp

r10s:
add some resilience against bad input parameter to netpgp.

may solve the problem with #333


r10s:
@muelli in  general, we commit patches to the upstream if we fix things here, eg. there are several delta chat patches accepted for libetpan.

wrt netpgp: we've done much more changes to netpgp than the ones of this pr. however, i do not know where there is an upstream repo that is maintained and would accept patches, any hints and help with doing upstream patches are welcome. however, might be that the most recent version is actually _here_.

PR #341, opened 2018-10-04
handle avatar image of user

r10s:
generally, we want the user to be able to set a profile image that is spreaded in some way to contacts. it is not yet decided how this should work in detail.

however, as a first step to this direction and as the android-dev ui needs a function to call just now, i've simply defined a setting that can be passed to `dc_set_config(context, "selfavatar", path)`.

the value is just an accessible path; the core makes a copy to the blob directory if needed and can use the file later, if required.

to get the avatar back, `dc_get_config(context, "selfavatar")`.

there is no need to adapt the bindings.

maybe this is sufficient, otherwise, we can refine this later as needed.

### DIFF
--- a/src/dc_context.c
+++ b/src/dc_context.c
@@ -213,9 +213,16 @@ void* dc_get_userdata(dc_context_t* context)
  * The function is called by dc_config_set*() and by dc_open().
  *
  * @private @memberof dc_context_t
+ * @param context The context object as created by dc_context_new().
+ * @param key Name of the value to update, NULL to update all.
+ * @return None.
  */
 static void update_config_cache(dc_context_t* context, const char* key)
 {
+	if (context==NULL) {
+		return;
+	}
+
 	if (key==NULL || strcmp(key, "e2ee_enabled")==0) {
 		context->e2ee_enabled = dc_sqlite3_get_config_int(context->sql, "e2ee_enabled", DC_E2EE_DEFAULT_ENABLED);
 	}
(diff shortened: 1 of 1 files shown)


src/dc_context.c
```diff
@@ -392,6 +399,9 @@ static char* get_sys_config_str(const char* key, const char* def)
  * - `server_flags` = IMAP-/SMTP-flags as a combination of @ref DC_LP flags, guessed if left out
  * - `displayname`  = Own name to use when sending messages.  MUAs are allowed to spread this way eg. using CC, defaults to empty
  * - `selfstatus`   = Own status to display eg. in email footers, defaults to a standard text
+ * - `selfavatar`   = File containing avatar.
+ *                    Will be copied to blob directory, if needed.
```
hpk42: what does the "if needed" mean?  I'd expect it to always get copied (and if it's there already, not copying is fine, but implementation detail optimization).

src/dc_context.c
```diff
@@ -413,9 +423,22 @@ int dc_set_config(dc_context_t* context, const char* key, const char* value)
 		return 0;
 	}
 
-	ret = dc_sqlite3_set_config(context->sql, key, value);
-	update_config_cache(context, key);
+	if (strcmp(key, "selfavatar")==0 && value)
+	{
+		char* rel_path = dc_strdup(value);
+		if (!dc_make_rel_and_copy(context, &rel_path)) {
+			goto cleanup;
```
hpk42: is relpath still free'd if we go to cleanup here?

src/dc_context.c
```diff
@@ -407,15 +416,29 @@ static char* get_sys_config_str(const char* key, const char* def)
...
+	{
+		rel_path = dc_strdup(value);
+		if (!dc_make_rel_and_copy(context, &rel_path)) {
+			goto cleanup;
+		}
+		ret = dc_sqlite3_set_config(context->sql, key, rel_path);
+	}
+	else
+	{
+		ret = dc_sqlite3_set_config(context->sql, key, value);
```
hpk42: not sure if it's harder in C wrt GC -- but in general i'd try to normalize this into only one dc_sqlite3_set_config call outside of the if-else-clauses.

r10s: in general, i agree. but in this case, is think the current code is more clear than having a separate `value_to_set` or so that must not be free()'d. we could also do a `value = rel_path`, but in general i don't like overwriting the given parameters very much (not saying i've not done this here and there)

Issue #343, opened 2018-10-04
add a function that shows the encryption state of the chat?


r10s:
if addition to `dc_msg_get_showpadlock()` we need sth. as `dc_chat_get_showpadlock()` to check out if a lock should be draw on the "send" button.

in older versions, `dc_msg_get_showpadlock()` imply that subsequent messages are also encrypted, however, this was not really understood by the users. 
also, this has the confusing disadvantage that the state of a messages changes if the user disables the e2ee-setting (as subsequent messages are no longer encrypted then ...).

hpk42:
in opportunistic chats there might be small dragons here UI-wise ... if i am starting to type a message and just before i hit "send" the send button changes from default-encrypt to default-not-encrypt because there was a new non-encrypted messages coming in -- it's a bit surprising/annoying maybe.  But it's an edge case, admittedly.   So maybe when "send" is actually clicked, it should tell the underlying code what to do (encrypt/not-encrypt) even though the opportunistic autocrypt states for the recipients might have changed meanwhile ...

r10s:
closing this for now, changes in this area should be driven by ui/ux, not necessarily function-wise.

Issue #342, opened 2018-10-04
get rid of dc_set_config_int() and dc_get_config_int()


r10s:
i would suggest to get rid of the functions `dc_set_config_int()` and `dc_get_config_int()`.

these functions are wrappers around the `dc_set_config()` and `dc_get_config()` functions, it's a c-thingie, where converting from/to strings is not that simple.

however, while deltachat-core-internal, this may make sense, for the bindings and the official api they raise more questions than bringing advantages - eg. what happens when i call set_int on a string. also internally, there are always two places to check and at least think about, when doing sth. special as converting abs/rel paths.

moreover, some binding do not have strong types (js, python, i think @hpk42 @ralphtheninja) and/or it is much easier to pass a numeric value than in c.

so, to keep things simple, i would suggest to mark them as "deprecated" and remove them some time later.

ralphtheninja:
> so, to keep things simple, i would suggest to mark them as "deprecated" and remove them some time later.

:+1: 

Sounds good!

hpk42:
fine with me -- but note that python does have "strong types" -- python will not automatically convert a string to an int or vice versa but simply raise a TypeError.

hpk42:
am generally all for keeping the c-core lean and mean ... much easier to convert/do things in java, js, python, ... compared to c.

hpk42:
On Thu, Oct 04, 2018 at 05:12 -0700, Lars-Magnus Skog wrote:
> > so, to keep things simple, i would suggest to mark them as "deprecated" and remove them some time later.
> 
> :+1: 
> 
> Sounds good!

IMHO it's fine to just strike set|get_config_int.
android/ios/desktop don't follow MASTER automatically so it's fine i think.

PR #344, opened 2018-10-04
remove settings cache, and cast-settings-functions

r10s:
1. remove settings cache as discussd in irc today
2. remove cast-settings functions as discussed in #342 

moreover, the dc_msg_get_padlock() now just returns the encryption-state of the message and does no longer imply that subsequent messages will be encrypted, too (was never really understood this way, imho). 
the latter should can be done by a button on the send-button or so.

iirc, i discussed this with several people some time ago, however, as 1. touches this function i think it is a good moment to start with this. for the send button, i created a new issue #343

### DIFF
--- a/src/dc_chat.c
+++ b/src/dc_chat.c
@@ -2035,9 +2035,13 @@ static uint32_t send_msg_raw(dc_context_t* context, dc_chat_t* chat, const dc_ms
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


r10s:
the tests fail because of the missing dc_get|set_config_int() functions.
if that's a problem, i can re-add them for now.

hpk42:
nope, just pushed a commit which should fix the branch.

Issue #337, opened 2018-10-01
'error IDEA is disable` on mac (Travis)


ralphtheninja:
After updating core in node bindings I get the following build error on Travis for mac:

```
[150/200] cc -Ilibs/netpgp/libs@netpgp@@netpgp@sta -Ilibs/netpgp -I../libs/netpgp -I../libs/netpgp/include -I/usr/local/Cellar/openssl/1.0.2p/include -Xclang -fcolor-diagnostics -pipe -Wall -Winvalid-pch -O0 -g -DHAVE_CONFIG_MESON -pthread  -MD -MQ 'libs/netpgp/libs@netpgp@@netpgp@sta/src_reader.c.o' -MF 'libs/netpgp/libs@netpgp@@netpgp@sta/src_reader.c.o.d' -o 'libs/netpgp/libs@netpgp@@netpgp@sta/src_reader.c.o' -c ../libs/netpgp/src/reader.c
FAILED: libs/netpgp/libs@netpgp@@netpgp@sta/src_reader.c.o 
cc -Ilibs/netpgp/libs@netpgp@@netpgp@sta -Ilibs/netpgp -I../libs/netpgp -I../libs/netpgp/include -I/usr/local/Cellar/openssl/1.0.2p/include -Xclang -fcolor-diagnostics -pipe -Wall -Winvalid-pch -O0 -g -DHAVE_CONFIG_MESON -pthread  -MD -MQ 'libs/netpgp/libs@netpgp@@netpgp@sta/src_reader.c.o' -MF 'libs/netpgp/libs@netpgp@@netpgp@sta/src_reader.c.o.d' -o 'libs/netpgp/libs@netpgp@@netpgp@sta/src_reader.c.o' -c ../libs/netpgp/src/reader.c
In file included from ../libs/netpgp/src/reader.c:92:
/usr/local/Cellar/openssl/1.0.2p/include/openssl/idea.h:65:4: error: IDEA is disabled.
#  error IDEA is disabled.
   ^
```

@flub Any idea?

Full build log here https://travis-ci.org/deltachat/deltachat-node/jobs/435493605

PR: https://github.com/deltachat/deltachat-node/pull/148

ralphtheninja:
[...]

PR #346, opened 2018-10-05
Try fixing the missing IDEA stuff on mac

flub:
@ralphtheninja if i'm right/lucky this could solve your issue #337, try it out.  And apologies for the delay looking at this.


ralphtheninja:
Pushed to this repo and rebased on latest master (due to meson.build error)

PR #361, opened 2018-10-08
remove DC_MSG_UNDEFINED from API

r10s:
just remove the DC_MSG_UNDEFINED viewtype from the API.

internally, 0=undefined is still used, but it is no longer valid for input from the API and is no longer returned.


ralphtheninja:
nice!

PR #375, opened 2018-10-09
make set_config|get_config throw an informative error if you try to g…

hpk42:
…et/set non-existing config keys (follow up on the merged #373 )


r10s:
lgtm

Issue #350, opened 2018-10-06
problems using forcefallback or monolithic builds


r10s:
playing around with the wrap-mode and monolith options (wanted to get a cmdline tool that uses the same openssl as the android version, to check out some things wrt slowness), i got the following errors:

- error using `meson -Dmonolith=true`   
`libs/netpgp/meson.build:59:11: ERROR:  Dependencies must be external dependencies`

- error using `meson --wrap-mode=forcefallback`  
`meson.build:44:2: ERROR:  fallback for openssl not found`

- error using `meson --wrap-mode=forcefallback --default-library=static`  
`libs/netpgp/meson.build:22:9: ERROR:  Dependencies must be external dependencies`

i am using meson 0.48.0 and a normal meson/ninja works like a charme.

might be that i am doing sth. completely wrong or that it's just my machine - but maybe @flub or someone else has an idea :)

flub:
I can at least reproduce the first issue.  Not tried the others yet.

So seems like I accidentally broke something, I'll investigate.

r10s:
@flub great, thank you. 
i think with very good chance, all three issues have the same reason :)

flub:
So 1 & 3 were a similar problem, 2 is simply not supported but you clipped the line that says that (a few lines higher)

flub:
Thanks for noticing btw!

r10s:
@flub great, thank you very much :)

PR #379, opened 2018-10-09
Build the lib in a few supported ways

flub:
I'm no travis expert and this doesn't really feel light the right way to do this, just chaining it onto the end.  So if someone has a better suggestion that'd be great too.

This will not pass until #378 is merged.  It would also have stopped #350 from happening.


flub:
The commits are all messed up, trying this again

PR #384, opened 2018-10-11
allow text added to images/files/etc.

r10s:
this pr makes deletachat-core able to add text to images/file/music/voice messages.

the text is added to the message object as usual using `dc_msg_set_text()`.

receiving such messages is also straight forward: _any_ message type can now contain text that can be retrieved using `dc_msg_get_text()`. 

this change makes it much easier to add some lines of text to a file.
the added text also appears smart together with the image/file in non-delta-muas. 

to allow the old telegram-ui to show these new compound messages without changes, i've added the function dc_no_compound_msgs() that let the core generate single messages for the description, however, this function is not documented and will be removed soon. all UIs should just be able to show the text attached to files.

btw. i do not think that more complex compound messages make much sense - they require a much more complex ui and also cause larger problems in flaky networks as there is only one "large" message to sent. just came into my mind as the flaky network discussion is currently ongoing :)

### DIFF
--- a/cmdline/cmdline.c
+++ b/cmdline/cmdline.c
@@ -449,7 +449,7 @@ char* dc_cmdline(dc_context_t* context, const char* cmdline)
 				"groupimage [<file>]\n"
 				"chatinfo\n"
 				"send <text>\n"
-				"sendimage <file>\n"
+				"sendimage <file> [<text>]\n"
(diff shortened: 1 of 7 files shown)


hpk42:
looks good to me although i can't claim to fully grasp the context of all the changes.

PR #385, opened 2018-10-11
copy files before dc_msg_send() returns

### LINKED ISSUE #374: (option to) copy files directly when dc_msg_set_file() is called.
currently, when you do set_file() you need to make sure that the file remains in place until the smtp-job gets around to transmit the thing.  Is there a particular reason to delay the copy to some uncertain future? Most of the time these files are <1MB and that should be very quick to copy on any device, no?

r10s:
this pr copies files before dc_msg_send() returns instead of postponing the copy action to the smtp-thread.

as files given to dc_msg_send() are no longer needed after dc_msg_send() returns, the event DC_EVENT_FILE_COPIED is no longer needed (eg. files can be deleted at once and without waiting for an event).

this change should make the overall usage of the sending message API easier and closes #374

### DIFF
--- a/src/dc_tools.c
+++ b/src/dc_tools.c
@@ -1431,10 +1431,10 @@ void dc_make_rel_path(dc_context_t* context, char** path)
 
 /**
  * Check if a path describes a file in the blob directory.
- * The path can be give absolute or relative (starting with `$BLOBDIR`).
+ * The path can be absolute or relative (starting with `$BLOBDIR`).
  * The function does not check if the file really exists.
  */
-int dc_is_in_blobdir(dc_context_t* context, const char* path)
+int dc_is_blobdir_path(dc_context_t* context, const char* path)
 {
 	if ((strncmp(path, context->blobdir, strlen(context->blobdir))==0)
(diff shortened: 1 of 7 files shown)


src/dc_tools.c
```diff
@@ -1429,6 +1429,21 @@ void dc_make_rel_path(dc_context_t* context, char** path)
 }
 
 
+/**
+ * Check if a path describes a file in the blob directory.
+ * The path can be give absolute or relative (starting with `$BLOBDIR`).
```
hpk42: strike the "give"

src/dc_tools.c
```diff
@@ -1429,6 +1429,21 @@ void dc_make_rel_path(dc_context_t* context, char** path)
...
+ * The function does not check if the file really exists.
```
hpk42: i suggest ``is_blobdir_path`` naming if the function does not check existenc.e

hpk42:
apart from some minor nitpicky comments this looks good -- +1 that the event is gone.  Some problems can be solved by removing a level of indirection :)

r10s:
thanks, @hpk42 , last commit target your comments :)

Issue #365, opened 2018-10-08
Introduce helper to get at "Received" mime headers for incoming messages


hpk42:
For performance debugging of the network messaging pipeline it'd be very handy to have a function that returns all ``Received`` headers and the ``Date`` header if possible, for a given incoming message.  
A possible naming is `dc_get_perf_mime_headers_incoming(msg_id)`.  This should return a multi-line string with at least the Received or Date headers (and maybe additional headers). For outgoing message this should return NULL. 

note that the `incoming` suffix is there because in the future we also might want to get Received-headers for "outgoing" messages by requesting them from the receiver which could probably send it back as part of the read-MDN.  Once this issue is resolved, another issue or PR could note/address the outgoing case. 

Even with just the "incoming" case covered we could have two bots sending each other messages through the same or two e-mail providers and measure the different hops.  this can be used for systematic provider-analysis given we have test accounts with them.

r10s:
@hpk42 just had a look at this.

there seems to be no easy way to get the raw headers from libetpan (i did not really expected this), however, i noticed that [there is already an undocumented debug option called "save_eml"](https://github.com/deltachat/deltachat-core/blob/master/src/dc_receive_imf.c#L1458): if `dc_set_config(context, "save_eml", "1")` was executed, all received headers and messages are written rawly to files. unfortunately, it is hard to guess from the msg_id to the filename, however, we could change this by naming the files as  `msg_id_<msgid>.eml`.

- the files are written to the blob-directory, so it should be easy to access them. 
- header parsing should not be more complicated as if one gets only the header
- the feature may become handy also in other situations
- using a switch to start saving these files might also be a good idea. this way, they are not saved by default and leave maybe unwanted traces on the user's device.
- using files instead of the database is also not that bad. esp. when it comes to the planned trimming of chats, it is needed to scan the blob directory for unused files; this way, the raw files may get deleted eg. after a day or so
- it is already there - and of course, we can tweak things later :)

what do you think?

hpk42:
On Mon, Oct 15, 2018 at 04:34 -0700, björn petersen wrote:
> @hpk42 just had a look at this.
> 
[...]

PR #388, opened 2018-10-15
add a switch and a function to get mime-headers

r10s:
- add a new function to get the mime-headers of messages, currently incoming only, dc_get_mime_headers()
- to collect the headers so that the function can word, one has to set the flag `save_mime_headers` using dc_set_config()

in contrast to my proposal in #365, i save the headers in a database as i do not want to mix files of too many different sources in the blob directory.

### DIFF
--- a/python/src/deltachat/account.py
+++ b/python/src/deltachat/account.py
@@ -2,6 +2,7 @@
 
 from __future__ import print_function
 import threading
+import six
 import re
 import time
 import requests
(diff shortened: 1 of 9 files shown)


src/dc_context.c
```diff
@@ -376,6 +376,7 @@ static char* get_sys_config_str(const char* key)
  * - `e2ee_enabled` = 0=no end-to-end-encryption, 1=prefer end-to-end-encryption (default)
  * - `mdns_enabled` = 0=do not send or request read receipts,
  *                    1=send and request read receipts
+ * - `save_mime_headers` = set this to "1" if you want to use dc_get_mime_headers() later
```
hpk42: wasn't there a list somewhere returned by ``sys.config_keys``?

r10s: f*ck ...

hpk42: you warned against it ;)

python/src/deltachat/account.py
```diff
@@ -222,6 +223,25 @@ def get_message_by_id(self, msg_id):
...
+
+        :param msg_id: integer message id
+        :returns: email-mime message object.
+        """
+        import email.parser
+        mime_headers = lib.dc_get_mime_headers(self._dc_context, msg_id)
+        if mime_headers:
+            s = ffi.string(mime_headers)
+            if isinstance(s, bytes):
+                s = s.decode("ascii")
```
hpk42: not sure if this is fully correct but it works in my setup. i think the char-string returned from dc_get_mime_headers is an ascii-string -- unicode chars in mime-header values have a special encoding (dealt with by the email parsing message object which this function returns)

hpk42:
not sure -- headers are a bit special -- they don't use "utf8" encoding
but some odd other encoding (would have to look up ...) that is expressed
with 7-bit chars.  This is all quite different from the encoding
of the mime-body/payload.

PR #390, opened 2018-10-16
cleanup some legacy code

r10s:
- remove searching for the old `X-Mr-*` prefix in mime headers (not send since ~1y, much longer that the changing of the api-function-names)
- remove the wrapper for the uuencoding functions, see https://github.com/deltachat/deltachat-core/issues/90#issuecomment-429527483 for some backgrounds


r10s:
good to have the travis build check, btw. 
would have forgotten to adapt the build files otherwise :)

hpk42:
nice :)

PR #392, opened 2018-10-16
remove legacy mrmailbox.h

### LINKED ISSUE #391: Remove mrmailbox.h
The mrmailbox.h definitions are not used from python, android or node. Once https://github.com/deltachat/deltachat-ios/issues/15 is done we can remove it from core.

r10s:
@jonasreinsch confirmed in https://github.com/deltachat/deltachat-ios/issues/15 that it is no longer neeed by ios.

closes #391


hpk42:
\o/

PR #393, opened 2018-10-16
remove deprecated dc_send_X_msg() functions

r10s:
- node/android does no longer use them
- ios has never used them (supports only images yet)


r10s:
i would suggest to keep `dc_send_text_msg()` just to have easy examples and for quick testing.

PR #396, opened 2018-10-16
remove untidy and incomplete mediainfo functions

r10s:
this pr removes the functions dc_msg_get_mediainfo() and dc_msg_set_mediainfo() that was meant to be used to set an author and a trackname for a media file.

however, i think these functions were a bit overambitious. also they are not well supported. eg. the information are written partly to the filename and also read from there as there is no easy way to pass them to a media file (id3 is too special).

also these functions were very special - they are only used when sending mp3 music files - might me that i was too close to my old silverjuke-project in the beginning :)

finally, only author and trackname are not even suffient, to do things right, we would need probably much more.

so - i would suggest to remove them. it seems to be totally fine to display the filename of mp3 files - as done for all other documents as well, so no special handling.

### DIFF
--- a/cmdline/stress.c
+++ b/cmdline/stress.c
@@ -470,8 +470,6 @@ void stress_functions(dc_context_t* context)
 		assert( DC_PARAM_HEIGHT == 'h' );
 		assert( DC_PARAM_DURATION == 'd' );
 		assert( DC_PARAM_MIMETYPE == 'm' );
-		assert( DC_PARAM_AUTHORNAME == 'N' );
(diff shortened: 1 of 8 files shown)


hpk42:
ah, the joy of removing code! :)

indeed i think that core should be rather dumb wrt to media files.  UIs can still offer to manipulate/specialized-view media files -- a bit like you can crop images, paint over them and then send ... maybe at some point there can be a doc/spec that says what Delta UIs are expected to be able to show ...

PR #407, opened 2018-10-24
steps towards single-folder milestone

### LINKED ISSUE #72: Delta-core to always refer to last message for improved non-Delta mail app threading
When i write a mail with my non-Delta mail app to Delta/Android and Delta/Android replies, it shows up in a totally different mail thread, i.e. does not reference my mail.  I suggest that Delta always references the last mail in a message chat window -- seen from the non-Delta mail app that feels natural and for Delta/Android it doesn't change anything: the message and its replies will simply appear in chronological order.

### LINKED ISSUE #370: Bcc-Self: send outgoing to self, remove "Sent" folder actions
Currently, each outgoing message is uploaded twice over the network:

- once per SMTP 
- once by uploading it into the Sent folder

considering that "uploads" are might be videos or larger attachments, it's a good idea to try to avoid this double-upload.  Also, chat-messages in the sent-folder are hardly useful.  UIs like Gmail always show sent-mails in the inbox conversations.  So we can just BCC our own address which will place the sent mail into our inbox.  Ideally, we'd then not IMAP-download this message (we know it's Message-Id so this should be possible in principle) but it's at least better than a double-upload because downloads are usually much faster.  

In any case, it means that we probably don't need to fetch anything from the Sent-folder which also eases the multi-device story as Delta/Desktop would see any outgoing messages automatically when it watches the INBOX and Delta/Android has sent a message (and vice versa).

### LINKED ISSUE #371: don't move messages to DeltaChat folder
The logic for moving messages from INBOX to DeltaChat is buggy and somewhat complex.  It's also unclear how multi-device (Desktop and Android or iOS) can easily remain synchronized.  We have a single IMAP thread and this can only watch exactly one IMAP folder through IMAP/Idle. It's therefore useful to try out not doing any such moves.  Probably a good idea to first implement better threading (#72).

### LINKED ISSUE #372: Make "INBOX" configurable
Allow to configure which "INBOX" is watched by DeltaChat.  Users may then add server-side filters to put all messages they want to have in DeltaChat (maybe to a particular alias) into a configured folder -- keeping a clean non-chat INBOX for other MUA.

r10s:
- tackle #370 by forcing SELF being always on the recipient list
- tackle #370 by completely removing the call to the IMAP upload
- #371 is also tackled as the messages that arrive in the INBOX now are not moved to DeltaChat folder. EDIT: by 0123a064f9c6208d987df4a0b4c83b2d3863ea9c this is also true for incoming messages
 
to keep the changes more clear, currently reformatting is skipped; this can be done later. as well as removing dead code.

messages are not appearing twice as there is already a check for the Message-ID in dc_receive_imf(), so, for now, this is fine. later on, we can optimize this by checking the Message-ID without completely downloading the whole message.

closes #370, closes #371, closes #372, closes #72

### DIFF
--- a/src/dc_chat.c
+++ b/src/dc_chat.c
@@ -2141,8 +2141,9 @@ static uint32_t send_msg_raw(dc_context_t* context, dc_chat_t* chat, const dc_ms
 	}
 	dc_param_set(msg->param, DC_PARAM_ERRONEOUS_E2EE, NULL); /* reset eg. on forwarding */
 
-	// setup In-Reply-To: and corresponding fields
-	// according to RFC 5322 3.6.4, page 25
+	// set "In-Reply-To:" to identify the message to which the composed message is a reply;
+	// set "References:" to identify the "thread" of the conversation;
+	// both according to RFC 5322 3.6.4, page 25
 	if (get_parent_mime_headers(chat, &parent_rfc724_mid, &parent_in_reply_to, &parent_references)) {
 		if (parent_rfc724_mid && parent_rfc724_mid[0]) {
 			new_in_reply_to = dc_strdup(parent_rfc724_mid);
--- a/src/dc_chat.c
+++ b/src/dc_chat.c
@@ -1997,6 +1997,7 @@ static int get_parent_mime_headers(const dc_chat_t* chat,
 		goto cleanup;
 	}
 
+	// use the last messsage of another user in the group as the parent
 	stmt = dc_sqlite3_prepare(chat->context->sql,
 		"SELECT rfc724_mid, mime_in_reply_to, mime_references"
 		" FROM msgs"
@@ -2008,6 +2009,24 @@ static int get_parent_mime_headers(const dc_chat_t* chat,
 		*parent_in_reply_to = dc_strdup((const char*)sqlite3_column_text(stmt, 1));
 		*parent_references  = dc_strdup((const char*)sqlite3_column_text(stmt, 2));
 		success = 1;
+	}
+	sqlite3_finalize(stmt);
+	stmt = NULL;
+
+	if (!success) {
+		// there are not messages of other users - use the first message if SELF as parent
+		stmt = dc_sqlite3_prepare(chat->context->sql,
+			"SELECT rfc724_mid, mime_in_reply_to, mime_references"
+			" FROM msgs"
+			" WHERE timestamp=(SELECT min(timestamp) FROM msgs WHERE chat_id=? AND from_id==?);");
+		sqlite3_bind_int  (stmt, 1, chat->id);
+		sqlite3_bind_int  (stmt, 2, DC_CONTACT_ID_SELF);
+		if (sqlite3_step(stmt)==SQLITE_ROW) {
+			*parent_rfc724_mid  = dc_strdup((const char*)sqlite3_column_text(stmt, 0));
+			*parent_in_reply_to = dc_strdup((const char*)sqlite3_column_text(stmt, 1));
+			*parent_references  = dc_strdup((const char*)sqlite3_column_text(stmt, 2));
+			success = 1;
+		}
 	}
 
 cleanup:
--- a/src/dc_chat.c
+++ b/src/dc_chat.c
@@ -1986,7 +1986,9 @@ static int last_msg_in_chat_encrypted(dc_sqlite3_t* sql, uint32_t chat_id)
 
 static uint32_t send_msg_raw(dc_context_t* context, dc_chat_t* chat, const dc_msg_t* msg, time_t timestamp)
 {
-	char*         rfc724_mid = NULL;
+	char*         new_rfc724_mid = NULL;
+	char*         new_references = NULL;
+	char*         new_in_reply_to = NULL;
 	sqlite3_stmt* stmt = NULL;
 	uint32_t      msg_id = 0;
 	uint32_t      to_id = 0;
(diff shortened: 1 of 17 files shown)


hpk42:
looks good -- it's a very interesting PR -- only containing removals, no additions! :+1:

r10s:
just noticed that SELF is added to `To:`, will fix that.

src/dc_chat.c
```diff
@@ -2102,10 +2104,49 @@ static uint32_t send_msg_raw(dc_context_t* context, dc_chat_t* chat, const dc_ms
...
+			new_references = dc_strdup(parent_references);
+		}
+		else if (parent_in_reply_to && parent_in_reply_to[0]
+		      && parent_rfc724_mid && parent_rfc724_mid[0]) {
+			new_references = dc_mprintf("%s %s", parent_in_reply_to, parent_rfc724_mid);
+		}
+		else if (parent_in_reply_to && parent_in_reply_to[0]) {
+			new_references = dc_strdup(parent_in_reply_to);
+		}
+	}
```
hpk42: i wonder what we need the references now for -- could you add a comment in the code on why this is needed?

src/dc_job.c
```diff
@@ -366,6 +320,12 @@ static void dc_job_do_DC_JOB_SEND_MSG_TO_SMTP(dc_context_t* context, dc_job_t* j
 			goto cleanup; /* unrecoverable */
 		}
 
+		/* add SELF to the recipient list (it's no longer used elsewhere, so a copy of the whole list is needless) */
+		if (clist_search_string_nocase(mimefactory.recipients_addr, mimefactory.from_addr)==0) {
+			clist_append(mimefactory.recipients_names, NULL);
+			clist_append(mimefactory.recipients_addr,  (void*)dc_strdup(mimefactory.from_addr));
+		}
+
```
hpk42: this is effectively a BCC, i.e. recipients are not added to the "To" or "CC" headers, right?

hpk42:
looks good to me on my first read.  Only wondering about if references are needed and if the recipients are showing in headers.

src/dc_job.c
r10s: exactly :)

src/dc_chat.c
r10s: a few lines above there was an, indeed, very short hint to the corresponding RFC; i added some more lines in  https://github.com/deltachat/deltachat-core/pull/407/commits/c7a4e76b09df261ae439f2aacac8aef6c0f6f071

src/dc_chat.c
```diff
@@ -1984,9 +1984,65 @@ static int last_msg_in_chat_encrypted(dc_sqlite3_t* sql, uint32_t chat_id)
...
+		*parent_rfc724_mid  = dc_strdup((const char*)sqlite3_column_text(stmt, 0));
+		*parent_in_reply_to = dc_strdup((const char*)sqlite3_column_text(stmt, 1));
+		*parent_references  = dc_strdup((const char*)sqlite3_column_text(stmt, 2));
+		success = 1;
+	}
+	sqlite3_finalize(stmt);
+	stmt = NULL;
+
+	if (!success) {
+		// there are not messages of other users - use the first message if SELF as parent
```
hpk42:
```suggestion
		// there are no messages of other users - use the first message if SELF as parent
```

src/dc_chat.c
```diff
@@ -1984,9 +1984,65 @@ static int last_msg_in_chat_encrypted(dc_sqlite3_t* sql, uint32_t chat_id)
...
+		stmt = dc_sqlite3_prepare(chat->context->sql,
+			"SELECT rfc724_mid, mime_in_reply_to, mime_references"
+			" FROM msgs"
+			" WHERE timestamp=(SELECT min(timestamp) FROM msgs WHERE chat_id=? AND from_id==?);");
```
hpk42: again i don't understand why this is constrained by from_id. probably missing something?

src/dc_receive_imf.c
```diff
@@ -1267,15 +1269,34 @@ void dc_receive_imf(dc_context_t* context, const char* imf_raw_not_terminated, s
 				}
 			}
 
+			if ((field=dc_mimeparser_lookup_field(mime_parser, "In-Reply-To"))!=NULL
+			 && field->fld_type==MAILIMF_FIELD_IN_REPLY_TO)
+			{
+				struct mailimf_in_reply_to* fld_in_reply_to = field->fld_data.fld_in_reply_to;
+				if (fld_in_reply_to) {
+					mime_in_reply_to = dc_str_from_clist(field->fld_data.fld_in_reply_to->mid_list, " ");
```
hpk42: 10-20 lines above you save mime headers -- is it correct to set mime_in_reply_to afterwards instead of before?

r10s: this should be fine.

- dc_str_from_clist() reads the header from the hashed-cache to find out the current value. no headers are modified here
- "10-20 lines above" we tear off the header from the original message to allow parsing and testing from python - this is also only a reading process

so, both will contain the same headers.
the writing to db happens a few lines later then.

src/dc_chat.c
```diff
@@ -1984,9 +1984,65 @@ static int last_msg_in_chat_encrypted(dc_sqlite3_t* sql, uint32_t chat_id)
...
+	if (chat==NULL
+	 || parent_rfc724_mid==NULL || parent_in_reply_to==NULL || parent_references==NULL) {
+		goto cleanup;
+	}
+
+	// use the last messsage of another user in the group as the parent
+	stmt = dc_sqlite3_prepare(chat->context->sql,
+		"SELECT rfc724_mid, mime_in_reply_to, mime_references"
+		" FROM msgs"
+		" WHERE timestamp=(SELECT max(timestamp) FROM msgs WHERE chat_id=? AND from_id!=?);");
```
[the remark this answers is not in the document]
r10s: the from_id is used here to attach answers to the last message of another user, not to the last message sent by SELF.
the from_id-condition only checks for SELF vs. NOT-SELF

my idea was to keep the thread as flat as possible if Bob adds several message to a message of Alice:

```
MsgAlice1
 |_ MsgBob1
 |_ MsgBob2
 |_ MsgBob3
```

Without checking the from_id, the thread would look like the following

```
MsgAlice1
 |_ MsgBob1
     |_ MsgBob2
         |_ MsgBob3
```

only if there are no foreign messages in the chat yet, answers are attached to SELF

src/dc_chat.c
r10s: see above.

src/dc_chat.c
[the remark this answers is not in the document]
hpk42: nice, ok -- makes sense.

PR #411, opened 2018-10-29
Simplify and unify string handling

### LINKED ISSUE #410: simplify string handling
currently, the string handling is a bit weird.

- [DC_EVENT_GET_QUANTITIY_STRING](https://c.delta.chat/group__DC__EVENT.html#gad7b9e617ad7cced53097a599c383fbc2) should replace returned placeholders as it is done by [DC_EVENT_GET_STRING](https://c.delta.chat/group__DC__EVENT.html#gadb561ea83f96d736cf53e38a23af20f1) (placeholders may be `%1$s` or `%2$s` or `%i` - we should also add some documentation about this)  
this way, implementations can offer a way to give the strings in advance and it is fine if the quantity parameter is ignored.

- EDIT: maybe we can also combine DC_EVENT_GET_QUANTITIY_STRING and DC_EVENT_GET_STRING to a single event where data1 is always the quantity and it's up to the implementation whether to make use of this information or not.

-  the DC_STR_* constants should be moved to deltachat.h

the changes would only result in very small changes in the core.

in general, i would suggest to leave the general approach of the core calling back the ui to get a string - caching the strings in advance may make language switches more complicated and also the quantity issue is already solved this way.  
with the changes above, moreover, it is rather simple to provide the strings in advance; we could provide a small helper class as we've already done for the [event queue](https://github.com/deltachat/deltachat-node/blob/master/src/eventqueue.c).

getting the strings completely out of the core looks also attractive at the first glance, however, this would make the things probably even more complicated as the core has to provide all information to build up strings, eg. also while error logging and so on.

related: https://github.com/deltachat/deltachat-desktop/issues/224

r10s:
- merge DC_STR_GET_STRING and DC_STR_GET_QUANTITY_STRING into DC_STR_GET_STRING(strId, quantity)
- move DC_STR_* constants to deltachat.h

closes #410

### DIFF
--- a/cmdline/main.c
+++ b/cmdline/main.c
@@ -52,7 +52,6 @@ static uintptr_t receive_event(dc_context_t* context, int event, uintptr_t data1
 	switch (event)
 	{
 		case DC_EVENT_GET_STRING:
-		case DC_EVENT_GET_QUANTITY_STRING:
 			break; /* do not show the event as this would fill the screen */
 
 		case DC_EVENT_INFO:
--- a/cmdline/stress.c
+++ b/cmdline/stress.c
@@ -581,7 +581,6 @@ void stress_functions(dc_context_t* context)
 		assert( !DC_EVENT_DATA2_IS_STRING(2010) );
 
 		assert(  DC_EVENT_RETURNS_STRING(2091) );
-		assert(  DC_EVENT_RETURNS_STRING(2092) );
 		assert(  DC_EVENT_RETURNS_STRING(2100) );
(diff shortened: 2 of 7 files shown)


r10s:
> Release v0.24.0 after this?

yes, good point :)

r10s:
@ralphtheninja shall i add the "version bump" to the pr or better bump it after merging?

ralphtheninja:
> @ralphtheninja shall i add the "version bump" to the pr or better bump it after merging?

I usually bump after merge. But up to you :)

Issue #403, opened 2018-10-18
Shift license to LGPL (or MPL?) 


hpk42:
There were some background discussions to change the license of the core and the python bindings. 
Idea is to allow also 3rd party independent commercial developments or other integrations that use  deltachat- core.  The slight preference with LGPL (compared to MPL) is that LGPL does not allow taking parts of the code and integrating it -- you must link/compile against it as a whole.  MPL allows taking parts of the source code and use it in a commercial product.  Through Deltacore we want to ensure compatibility between all apps that use it -- so LGPL's restriction seems more fitting than MPL in that respect.  Maybe @jankass can take a look at other potentially interesting lgpl/mpl differences but overall let's try to avoid long-drawn discussions about it -- if need be we can re-license again later i guess :) 

Looking at https://github.com/deltachat/deltachat-core/graphs/contributors
we need to get an "ok" from those who contributed significantly (changing a typo or changing two lines is not significant in the legal copyright sense).  If we just consider the C source code that'd be less people (some mostly contributed to packaging/building/glue).  But if we get an ok from the following mentioned we are safe and don't need to differentiate much. 

That's @r10s , me, probably @Ampli-fier @t-oster @flub @ralphtheninja @azul -- note that  @r10s and @azul were doing changes while employed with merlinux GmbH which in turn is fine with re-licensing to either LGPL or MPL (should we decide to rather go for that).

t-oster:
fine by me

ralphtheninja:
It's still free software. I don't mind at all.

r10s:
also fine by me.

Ampli-fier:
No objections from my side.

hpk42:
thanks all for answering! 

on freenode @flub brought up the iOS question -- what if someone wants to use deltachat-core in an iOS app. Indeed it seems LGPL with ios could be problematic https://stackoverflow.com/questions/459833/which-open-source-licenses-are-compatible-with-the-apples-iphone-and-its-offici#9070734

maybe that means we go for MPL after all ... seems less problematic.  (no need to again agree, as that was already part of the issue question).

flub:
For the record here: I've also been allowed to approve this by my employer.

PR #415, opened 2018-11-01
move to Mozilla Public License 2.0 and clarify overall licensing situation

hpk42:
- src, docs, python, cmdline is licensed under the MPL 2.0 
- libs/* contain 3rd party docs, licensed under their own respective licenses
- boilerplate for ".c" and ".h" files was removed. 
- the toplevel "LICENSE" file also has a note on Copyright and contributors. 

fixes https://github.com/deltachat/deltachat-core/issues/403


flub:
githug?  is that a thing?

LICENSE
```diff
@@ -1,19 +1,380 @@
 
-For licensing information of the libdeltachat library
-please see src/LICENSE 
+The files in this directory and under the "src", "python", "doc" and "cmdline" subdirectories
```
flub: negative matching ("all but the libs directory") might be easier?  not that i mind either way

PR #416, opened 2018-11-01
Build the core lib in a few supported ways

flub:


flub:
I have no idea why travis is not building this :(

hpk42:
thanks, this is nice and gives more certainty that we don't break a builld variant accidentally.

PR #417, opened 2018-11-01
use exponential backoff

### LINKED ISSUE #265: Messages sending may fail on poor network connection
@MMurzik reported at https://github.com/deltachat/deltachat-android/issues/386

> Before version 0.20.0 if a message was sent and connection to the mail server broke due to poor internet connectivity, Delta Chat would be sending this message over and over again, so it would definitely be delivered sooner or later, when connectivity improves.
Now, with the new "function" in version 0.20.0 to give up sending after a number of retries, reliable delivery of messages on poor connection is no more possible. When the mail server is inaccessible for some time (e.g. on a subway, or at the countryside, or even in flight mode), Delta Chat simply gives up sending these messages. That makes me constantly re-check if all messages were sent, and if not re-send it manually. It is inconvinient. And before version 0.20.0 I can simply send a message, forget about it, and to be sure that it will be delivered 100%
Please add option in settings to restore old behavior of Delta Chat to try sending unsent messages indefinite number of times. Thank you.

i moved this here as it is related to deltachat-core, not android.

the reason for not just "endless trying over and over" was avoiding excessive data consumption. in general, Delta Chat _should_ retry if it fails and the network is in "disconnected" state. however, of course there may be bugs and this may need some more research :)

related: [Size limit ends in endless upload #102](https://github.com/deltachat/deltachat-android/issues/102#issuecomment-404325534)

r10s:
make the retries on failures more reliable, see https://en.wikipedia.org/wiki/Exponential_backoff.

- the calculation of the backoff times is done in [get_backoff_time_offset()](https://github.com/deltachat/deltachat-core/pull/417/files#diff-3f291e506d02f02b7f7e2e9ba43234b4R413) which is based on the [playground-python-snippet](https://github.com/deltachat/playground/blob/master/expbackoff/compute.py)
- we use a MULTIPLY of 5 which results in random retries in the following intervals in seconds: [0..5] [0..15] [0..35] [0..75] [0..155] [0..315] [0..635] [0..1275] [0..2555] [0..5115] [0..10235] [0..20475] [0..40955] [0..81915] [0..163835] (the last one is about max. 2 days, we stop after 16 retries)
- we introduce a new function (o/ @ralphtheninja) called [dc_maybe_network()](https://github.com/deltachat/deltachat-core/pull/417/files#diff-3f291e506d02f02b7f7e2e9ba43234b4R992) which which can be called by the ui when there are reasons to believe network is available again. after this function is called, the core tries to get jobs done at once and bypasses the calculated time. on failure, however, the times are not reset but kept.  
it's fine if the ui _never_ calls this function, then everything is handled by the backoff times, which, however, will take more time

@hpk42 the calculated time-offsets are added to the starting time of the job currently. as this offset can be 0 at any interval, the new calculated started time may be in the _past_. in this case, the job is retried at once. this is technically no problem, however, i am not sure if this matches the initial idea. want just to mention this :)

btw: no idea why travis complains on this pr.

closes #265

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -35,6 +35,9 @@ endif
 # pthreads is not a real dependency
 pthreads = dependency('threads')
 
+# declare option to link against libm
+math = declare_dependency(link_args: ['-lm'])
+
 # zlib should move grow static-pic-lib support and be handled like
 # this as well.
 zlib = dependency('zlib', fallback: ['zlib', 'zlib_dep'])
--- a/meson.build
+++ b/meson.build
@@ -36,7 +36,8 @@ endif
 pthreads = dependency('threads')
 
 # declare option to link against libm
-math = declare_dependency(link_args: ['-lm'])
+cc = meson.get_compiler('c')
+math = cc.find_library('m')
 
 # zlib should move grow static-pic-lib support and be handled like
 # this as well.
--- a/cmdline/cmdline.c
+++ b/cmdline/cmdline.c
@@ -411,7 +411,7 @@ char* dc_cmdline(dc_context_t* context, const char* cmdline)
 				"configure\n"
 				"connect\n"
 				"disconnect\n"
-				"poll\n"
+				"maybenetwork\n"
 				"help imex (Import/Export)\n"
 				"==============================Chat commands==\n"
 				"listchats [<query>]\n"
@@ -638,6 +638,11 @@ char* dc_cmdline(dc_context_t* context, const char* cmdline)
 		if (ret == NULL) {
 			ret = COMMAND_FAILED;
 		}
+	}
+	else if (strcmp(cmd, "maybenetwork")==0)
+	{
+		dc_maybe_network(context);
+		ret = COMMAND_SUCCEEDED;
 	}
 
 	/*******************************************************************************
--- a/src/dc_context.h
+++ b/src/dc_context.h
@@ -50,6 +50,7 @@ struct _dc_context
 	dc_imap_t*       imap;                  /**< Internal IMAP object, never NULL */
 	pthread_mutex_t  imapidle_condmutex;
 	int              perform_imap_jobs_needed;
+	int              probe_imap_network;    /**< if this flag is set, the imap-job timeouts are bypassed and messages are sent until they fail */
 
 	dc_smtp_t*       smtp;                  /**< Internal SMTP object, never NULL */
 	pthread_cond_t   smtpidle_cond;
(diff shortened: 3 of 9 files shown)


testbird:
Is there a timer reset for quick re-conntection when the network is switched (provider or wlan network changes)?

r10s:
> Is there a timer reset for quick re-conntection when the network is switched (provider or wlan network changes)?

yes. the ui can call `dc_maybe_network()` if there are hints that the network changed/works again.

deltachat-core tries to send out all pending messages immediately in this case, in the order of their backoff times. 

however, on failure, the backoff times are not resetted (i think this would result in heavy load on flaky networks). but the ui can call `dc_maybe_network()` again every time to let the core try over.

hpk42:
line 2827 of https://travis-ci.org/deltachat/deltachat-core/jobs/453739642 says "undefined reference to `pow`

r10s:
> says "undefined reference to pow"

ahh, i see. there is the `-lm` linker flag for the math library missing.

have to figure out, how to set this with meson...

meson.build
```diff
@@ -35,6 +35,9 @@ endif
 # pthreads is not a real dependency
 pthreads = dependency('threads')
 
+# declare option to link against libm
+math = declare_dependency(link_args: ['-lm'])
```
flub: I think it's better to use:
```
cc = meson.get_compiler('c')
math = cc.find_library('m')
```

r10s: okay, thank you very much, i will change this.

hpk42:
not sure that a multiple of 5 is a good choice, would more think a multiple of 60.  Network changes will also lead to retrying so i think it's ok to try less often independently of network changes and do fewer than 16 retries (wouldn't mind more than 2 days though -- after all a mobile can be offline for a few days sometimes, and the retry-logic would always fail all messages then, if i see it correctly).

testbird:
The "exponential backoff" algorithm that you have implemented seems to *randomize* delays, as such it was originally intended to solve ethernet frame collisions, and not to wait until connection re-establishment. 

Don't you think removing the randomization would make the delays, retries and job execution order  more reliable?

Also, as said under "termination  (network disconnection events)" in in https://support.delta.chat/t/wiki-imap-strategy-to-chat-within-the-email-ecosystem/123 I think retries should not be done while there is no network, because doing so is useless and will not make the delay pattern work as intended after a reconnect.

Here is an article that explains this as "circuit breaker"
https://dzone.com/articles/understanding-retry-pattern-with-exponential-back

And this here seems to suggest to rather increase the protocol timeout for high-packet-loss networks (e.g. cellular, or wi-fi near the limits of its range):
https://stackoverflow.com/questions/38980577/adjusting-http-timeout-versus-backoff-during-retries

Also remember the other take from the original issues was, that jobs with *permanent errors* would need to to be stopped completely (e.g. retries to send too large videos wasted bandwith)? That seems not to be tracked anymore.

testbird:
BTW: I would limit the max. delay to 10 hours for transient errors, to ensure some daily retries at different times of the day.

testbird:
Even 10h looks to much for a chat app. 1-3h?

Issue #420, opened 2018-11-05
check key name on dc_set_config()


r10s:
[dc_set_config()](https://c.delta.chat/classdc__context__t.html#aff3b894f6cfca46cab5248fdffdf083d) should check that the first parameter (key name) is one of the ones returned by [dc_get_config("sys.config_keys")](https://c.delta.chat/classdc__context__t.html#a8876ed2c7ca6c54eac9c31bcc304b6ad)

same probably for dc_get_config().

cc @Jikstra

r10s:
closed by #421

PR #421, opened 2018-11-05
Check key name on dc_set_config() and dc_get_config()

bpiddubnyi:
#420

### DIFF
--- a/src/dc_context.c
+++ b/src/dc_context.c
@@ -367,14 +367,14 @@ static char* get_config_keys_str()
 	dc_strbuilder_init(&ret, 0);
 
 	for (int i = 0; i < str_array_len(config_keys); i++) {
-		if (ret.buf != NULL) {
+		if (strlen(ret.buf) > 0) {
 			dc_strbuilder_cat(&ret, " ");
 		}
 		dc_strbuilder_cat(&ret, config_keys[i]);
 	}
 
 	for (int i = 0; i < str_array_len(sys_config_keys); i++) {
-		if (ret.buf != NULL) {
+		if (strlen(ret.buf) > 0) {
 			dc_strbuilder_cat(&ret, " ");
 		}
 		dc_strbuilder_cat(&ret, sys_config_keys[i]);
--- a/src/dc_context.c
+++ b/src/dc_context.c
@@ -358,20 +358,24 @@ static int is_gettable_config_key(const char* key)
 }
 
 
-static char* get_config_keys_str() {
+static char* get_config_keys_str()
+{
 	dc_strbuilder_t  ret;
 	dc_strbuilder_init(&ret, 0);
 
 	for (int i = 0; i < str_array_len(config_keys); i++) {
+		if (ret.buf != NULL) {
+			dc_strbuilder_cat(&ret, " ");
+		}
 		dc_strbuilder_cat(&ret, config_keys[i]);
-		dc_strbuilder_cat(&ret, " ");
 	}
 
-	for (int i = 0; i < str_array_len(sys_config_keys) - 1; i++) {
+	for (int i = 0; i < str_array_len(sys_config_keys); i++) {
+		if (ret.buf != NULL) {
+			dc_strbuilder_cat(&ret, " ");
+		}
 		dc_strbuilder_cat(&ret, sys_config_keys[i]);
-		dc_strbuilder_cat(&ret, " ");
 	}
-	dc_strbuilder_cat(&ret, sys_config_keys[str_array_len(sys_config_keys) - 1]);
 
 	return ret.buf;
 }
--- a/src/dc_context.c
+++ b/src/dc_context.c
@@ -15,6 +15,35 @@
 #include "dc_apeerstate.h"
 
 
+static const char* config_keys[] = {
+	"addr",
+	"mail_server",
+	"mail_user",
+	"mail_pw",
+	"mail_port"
+	"send_server",
+	"send_user",
+	"send_pw",
+	"send_port"
+	"server_flags",
+	"displayname"
+	"selfstatus",
+	"selfavatar",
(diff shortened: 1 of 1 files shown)


bpiddubnyi:
As far as I can see, the tests are failing because config keys designed for the internal use by the `dc_sqlite3_{get,set}_config{,_int}()` only (`configured_addr`, `configured_mail_pw`, `configured`), are set with the public `dc_set_config()` function for the test purposes. https://github.com/deltachat/deltachat-core/blob/master/python/tests/conftest.py#L74

I see two possible solutions. First, is to create bindings for the `dc_sqlite*` functions. The second one is to allow user to set these keys (which, I beleive, is quite undesirable).

Also, hi, my name is Borys, and I would really like to help with the deltachat-core development.

ralphtheninja:
> Also, hi, my name is Borys, and I would really like to help with the deltachat-core development.

@bpiddubnyi :wave: I'm Magnus.

src/dc_context.c
```diff
@@ -307,6 +336,45 @@ char* dc_get_blobdir(const dc_context_t* context)
...
+	for (int i = 0; i < str_array_len(sys_config_keys); i++) {
+		if (strcmp(key, sys_config_keys[i]) == 0) {
+			return 1;
+		}
+	}
+	return is_settable_config_key(key);
+}
+
+
+static char* get_config_keys_str() {
```
r10s: just nitpicking:
linebreak missing (there is no concrete style guide, however, typically we use linebreaks after function names to have a clearer separation from for/while/if

src/dc_context.c
```diff
@@ -15,6 +15,35 @@
 #include "dc_apeerstate.h"
 
 
+static const char* config_keys[] = {
```
r10s: great!
much better to use an array here!

src/dc_context.c
```diff
@@ -320,14 +388,7 @@ static char* get_sys_config_str(const char* key)
...
 	{
-		return dc_strdup("addr"
-		       " mail_server mail_user mail_pw mail_port"
-		       " send_server send_user send_pw send_port"
-		       " server_flags displayname"
-		       " selfstatus selfavatar"
-		       " e2ee_enabled mdns_enabled"
-		       " save_mime_headers"
-		       " sys.version sys.msgsize_max_recommended sys.config_keys");
+		return get_config_keys_str();
```
r10s: again, much better!

src/dc_context.c
```diff
@@ -307,6 +336,45 @@ char* dc_get_blobdir(const dc_context_t* context)
...
+	for (int i = 0; i < str_array_len(config_keys); i++) {
+		dc_strbuilder_cat(&ret, config_keys[i]);
+		dc_strbuilder_cat(&ret, " ");
+	}
+
+	for (int i = 0; i < str_array_len(sys_config_keys) - 1; i++) {
+		dc_strbuilder_cat(&ret, sys_config_keys[i]);
+		dc_strbuilder_cat(&ret, " ");
+	}
+	dc_strbuilder_cat(&ret, sys_config_keys[str_array_len(sys_config_keys) - 1]);
```
r10s: just nitpicking:

i think, this is to avoid the last space.
if, for any reasons, sys_config_keys is empty in the future, this will cause reading invalid memory. currently, this will happen never, but maybe it is just easier to add all items in the loop and add the space only by checking `ret.buf[0]!=0`in both loops.

@bpiddubnyi what do you think? makes sense?

r10s:
> I see two possible solutions. First, is to create bindings for the dc_sqlite* functions. The second one is to allow user to set these keys (which, I beleive, is quite undesirable).

well, python uses this for testing, so, maybe it is okay and the easiest to just allow these keys.
we could also document this and say it is okay to set these keys if you do not want to call dc_configure() and if the user is really sure, the parameters are correct, eg. if he has tested them in another way.

src/dc_context.c
```diff
@@ -15,6 +15,35 @@
...
+	"send_pw",
+	"send_port"
+	"server_flags",
+	"displayname"
+	"selfstatus",
+	"selfavatar",
+	"e2ee_enabled",
+	"mdns_enabled",
+	"save_mime_headers",
+};
```
r10s: as mentioned in the additional comment, maybe we should add the `configured*` keys here

src/dc_context.c
```diff
@@ -307,6 +339,49 @@ char* dc_get_blobdir(const dc_context_t* context)
...
+}
+
+
+static char* get_config_keys_str()
+{
+	dc_strbuilder_t  ret;
+	dc_strbuilder_init(&ret, 0);
+
+	for (int i = 0; i < str_array_len(config_keys); i++) {
+		if (ret.buf != NULL) {
```
r10s: it must read `ret.buf[0]!=0` or `strlen(ret.buf)>0` as `dc_strbuilder_init()` always sets the pointer to an empty string.
the condition  is always `ret.buf != NULL` is always true therefore.

sorry for the bad or missing internal documentation.

src/dc_context.c
```diff
@@ -307,6 +339,49 @@ char* dc_get_blobdir(const dc_context_t* context)
...
+			dc_strbuilder_cat(&ret, " ");
+		}
+		dc_strbuilder_cat(&ret, config_keys[i]);
+	}
+
+	for (int i = 0; i < str_array_len(sys_config_keys); i++) {
+		if (ret.buf != NULL) {
```
r10s: same here.

src/dc_context.c
bpiddubnyi: oh, my bad. I actually saw this, but somehow immediately forgot even after you mentioned the possible solution in the comments. will fix in a couple of minutes.

bpiddubnyi: Done

PR #425, opened 2018-11-07
make dc_get_fresh_msgs() more reliable

r10s:
- do not return special chat_ids to the list (this was not notices in the old telegram-ui as the list was compared against the messages arrived through DC_EVENT_INCOMING_MSG, however, what should not be necessary)
- do not return NULL on errors, instead, return the empty list as also done for other functions
- add a bit more of documentation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ script:
 deploy:
   provider: script
   skip_cleanup: true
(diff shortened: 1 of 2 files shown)


hpk42:
thanks @r10s for the PR and @azul for the help with fixing the travis issue that was introduced by the matrix build.

PR #429, opened 2018-11-10
README: LibEtPan must be bullt with SASL support

steils:


ralphtheninja:
> one thing, that is still currently unclear to me for what exact reason sasl is needed at all :)

@r10s Is there an issue on this?

steils:
@r10s The exact reason is using `mailsmtp_auth()` in dc_smtp.c.
Then the internal funtion in libetpan mailesmtp_auth_sasl() is used, an if `USE_SASL` is not defined, I get "Not implemented" error message on SMTP-login.

ralphtheninja:
Can we get rid of it somehow? :smile:

r10s:
> @r10s Is there an issue for this?

now, there is one :)
https://github.com/deltachat/deltachat-core/issues/430

PR #433, opened 2018-11-14
add missing config-keys to positive list

r10s:
dc_get|set_config() checks for valid keys since 0.25.0; this pr adds the "configured" keys to the positive list.

normally, you should not _set_ them, but for special purposes and if you know what you're doing, it's fine.
a _get_ is always fine.


src/dc_context.c
```diff
@@ -34,7 +34,15 @@ static const char* config_keys[] = {
 	"mdns_enabled",
 	"save_mime_headers",
 	"configured_addr",
+	"configured_mail_server",
+	"configured_mail_user",
 	"configured_mail_pw",
+	"configured_mail_port"
```
ralphtheninja: Rebase needed and some missing commas.

r10s: currently working on this :)

PR #435, opened 2018-11-14
keep sort order if system time lags (do not sort before last message)

testbird:
a copy-paste attempt for https://github.com/deltachat/deltachat-android/issues/398 to review

* getting last_msg_time outside of is_fresh_msg
* use it to limit truncating

### DIFF
--- a/src/dc_receive_imf.c
+++ b/src/dc_receive_imf.c
@@ -259,42 +259,57 @@ static void calc_timestamps(dc_context_t* context, uint32_t chat_id, uint32_t fr
                                time_t* sort_timestamp, time_t* sent_timestamp, time_t* rcvd_timestamp)
 {
 	*rcvd_timestamp = time(NULL);
-
 	*sent_timestamp = message_timestamp;
-	if (*sent_timestamp > *rcvd_timestamp /* no sending times in the future */) {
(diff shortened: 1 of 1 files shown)


testbird:
> we could just remove the is_fresh_msg check from calc_timestamps() and we would have the same effect as you described above.

> bunch of others and/or make things more complicated

Well, but in which direction do you see the problems?
How should *limiting* the current truncating that happens with a resetted system time make things worse than before?

Because I tried to think of sequences of different system and message times but found these should all create subsequent sort_timestamps, even if those may actually still be in the past it seemed better than placing them before last_msg_time.



The build error I guess comes from my false placement of sqlite3_finalize() or variable scope?

testbird:
If you can explain that limiting the truncing *really* has a bad effect on imports, the lower bound could be ensured only for fresh messages.

testbird:
It should be fine to only have the bound when adding fresh messages.

testbird:
The reason I put the bounding behind the time smearing (outside of the is_fresh clause), however, is because the unconditional time smearing at the end also does the truncating of future times, without bounding it with the last message time.

My example just corrected the truncating afterwards, because otherwise the time smearing function may for example have to get support to get passed a bounding parameter.

Issue #293, opened 2018-09-19
refined event emission


hpk42:
Currently delta-core emits useful information (for testing and i think also for UIs/bindings) as ``DC_EVENT_INFO`` callback items.  Also it emits all kinds of debugging information you only sometimes need (smtp/connectivity checking etc.) as the same event. 

For example, regarding messages I suggest as new events in the format eventname, data1, data2:

- ``DC_MSG_SEEN <msg_id> <info message>``
- ``DC_MSG_READ <msg_id> <info message>``
- ``DC_MSGS_DELETED <array[msg_id]> <info message>``
- (more message related ones i guess) 

- ``DC_IMAP_DEBUG <> <info message>`` the several imap connection related infos``
- ``DC_SMTP_DEBUG <> <info message>`` the several smtp connection related infos ``
- ``DC_IMAP_IS_OPERATIVE <emailadr> <connection-string-with settings>`` this means, ssl-connection, and login was successful, we can execute  imap commands. 
- ``DC_SMTP_IS_OPERATIVE <emailadr> <connection-string-with settings>`` this means, ssl-connection and login was successful, we can send mails over smtp 

These all can mostly just replace existing DC_INFO messages, using the same c-machinery. 
I suggest to introduce these renamings step by step, along with python tests so we not only document their behaviour/emission but also ensure they continously work.

r10s:
> I guess it's also ok to stick to the DC_EVENT_ prefix for all of the mentioned events.

i would leave the prefix as otherwise there may be easy confusion with related flags, types and so on.
also it is easier to write a script that greps for all events to put them to other language defines.

PR #438, opened 2018-11-15
remove DC_EVENT_IS_OFFLINE

### LINKED ISSUE #426: do not rely on DC_EVENT_IS_OFFLINE
maybe we should not rely on DC_EVENT_IS_OFFLINE too much.

doing a grep, it is used on some critical places to restart things or not, eg. in dc_perform_imap_fetch().

i think we should have a critical look at the use of these calls and use DC_EVENT_IS_OFFLINE not in critical situations as, on flaky networks, checking the network status this way may not be a good friend.

may be related to #363 

we should target this after the [exponential-backup-pr](https://github.com/deltachat/deltachat-core/pull/417) (currently still WIP) is done.

r10s:
this pr stops using DC_EVENT_IS_OFFLINE to decide if to connect or to retry a job.

instead, the network action is just performed and if it fails, well, we know that.

there is one remaining usage of DC_EVENT_IS_OFFLINE: _if_ an network error occurred, only the first one is reported as an error that should be shown to the user (via DC_EVENT_ERROR) - if there is not network in that moment we say just "No Network" to the user instead of sth. as "Login with password xyz" failed as the latter is probably misleading. however, the latter is still logged as a warning.

closes #426
closes https://github.com/deltachat/deltachat-android-ii/issues/127
also tackles #293 a bit

### DIFF
--- a/src/dc_configure.c
+++ b/src/dc_configure.c
@@ -426,13 +426,6 @@ void dc_job_do_DC_JOB_CONFIGURE_IMAP(dc_context_t* context, dc_job_t* job)
 
 	PROGRESS(0)
 
-	if (context->cb(context, DC_EVENT_IS_OFFLINE, 0, 0)!=0) {
-		dc_log_error(context, DC_ERROR_NO_NETWORK, NULL);
-		goto cleanup;
-	}
-
-	PROGRESS(100)
-
 	/* 1.  Load the parameters and check email-address and password
 	 **************************************************************************/
 
--- a/src/dc_context.h
+++ b/src/dc_context.h
@@ -89,10 +89,10 @@ struct _dc_context
 };
 
 void            dc_log_error         (dc_context_t*, int code, const char* msg, ...);
-void            dc_log_error_if      (int* condition, dc_context_t*, int code, const char* msg, ...);
 void            dc_log_warning       (dc_context_t*, int code, const char* msg, ...);
 void            dc_log_info          (dc_context_t*, int code, const char* msg, ...);
-void            dc_log_event         (dc_context_t* context, int event_code, int code, const char* msg, ...);
+void            dc_log_event         (dc_context_t*, int event_code, int code, const char* msg, ...);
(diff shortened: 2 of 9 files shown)


hpk42:
hum, why not leave the precise message to be displayed to the UI caller?  If core generated a DC_EVENT_ERROR_NETWORK("login with password xyz failed") the UI can decide if it rather displays this as "network" error as we generally say the UI is best suited to determine if network is available.  I think it'd be better if we can get completely rid of DC_EVENT_IS_OFFLINE -- one callback and complexity gone.

ralphtheninja:
> there is one remaining usage of DC_EVENT_IS_OFFLINE: _if_ an network error occurred, only the first one is reported as an error that should be shown to the user (via DC_EVENT_ERROR)

Does this change mean we can remove code handling the event in e.g. the node bindings?

r10s:
> hum, why not leave the precise message to be displayed to the UI caller?

thought, we test first if DC_EVENT_IS_OFFLINE is needed or not.
but, thinking it over, i am pretty sure it is not needed for connectivity - in fact, android-dev has not implemented it yet and it works anyway :)

my initial idea was to move this decision away from the ui - and as there was DC_EVENT_IS_OFFLINE for other reasons it came at no cost for the ui - so, yes, it might be a good idea to get rid of this now.

in fact, there is even more logic in the error handling on the core side:

currently, the first "No network" event is reported as an error and subsequent "No network" events are reported as warnings - otherwise, on connection retries the "No network" bubble pops up again and again... (we reset the warning to an error when the user tries sending a message or if there _is_ network again)
This works fine for the view of the ui as the doc say, ERRORS should be shown to the user, WARNINGS should only be logged.
if we now use DC_EVENT_ERROR_NETWORK("bla") instead of DC_EVENT_ERROR|WARNING("no network), the ui has to do this on it's own, what might be rather complicated on the ui-side then.

suggestion: maybe pass a flag _is-subsequent-error_ together with the DC_EVENT_NETWORK_ERROR?

> Does this change mean we can remove code handling the event in e.g. the node bindings?

if we implement the suggestion from @hpk42  - yes :)

PR #439, opened 2018-11-15
adapt to new event names

hpk42:


hpk42:
thanks, feel free to merge in this case ;)

r10s:
> thanks, feel free to merge in this case ;)

next time then. so, if there is no WIP in the title, things are free to merge?

Issue #193, opened 2018-06-17
14 minute periodic IMAP-idle noop() calls?


testbird:
NOOP may be cheaper than frequent re-starting of a new idle watch on the server/client,
and network interruptions should be handled better and more quickly by listening to network events (instead of frequent IDLE restarts).

I have searched the repo but was not able to find a NOOP loop. 
I found the requirement for 14 minute timeouts mentioned at https://www.isode.com/whitepapers/imap-idle.html. It says:

> 
> In practice things are made more complex by the problem of timeouts occuring when there is no activity keeping the connection open. The main timeouts that will occur are:
> 
>   * IMAP server timeout: Typically occurs after 30 minutes with no activity.
>   * NAT Gateway timeout: Most mobile devices access the Internet through a device operated by the mobile service provider called a NAT (Network Address Translation) gateway. These will typically time out an idle connection after 15 minutes.
> 
> The solution to this is for the IMAP client to issue a NOOP (No Operation) command at intervals, typically every 15 minutes. This will exchange a few bytes of data, and keep everything active.

t-oster:
Shouln' the IDLE command be send every few minutes anyway? (It should usually be resent every 28min as I see, but right now it should be sent every minute https://github.com/deltachat/deltachat-core/blob/master/src/mrimap.c#L921)

testbird:
Thanks you, for pointing me to it. 

So I see in the released version 0.17.3 the IDLE_DELAY_SECONDS  is (28*60), and it only got later reduced to 23 and then 1 minute.

The intervals > 15 minutes in the release explains that the NAT timeout breaks the tcp connection if there is no activity at all. Consequently the idle thread only gets re-established after 30 minutes.

The current 1 minute polling during imap-idle mode may prevent NAT resets, but IIUC it's just a ugly workaround until smtp is in a separate thread, right? It somehow defeats the purpose of IDLE.


I found this, with an answer saying how messages should actually not get lost: 
https://stackoverflow.com/questions/2513194/imap-idle-timeout
Together with
https://joshdata.wordpress.com/2014/08/09/how-bad-is-imap-idle/
what I take away is:

* The imap-idle connections needs to be refreshed after 29 minutes, even if the client IP has not changed. Possibly start a new connection before closing the old one, in case some server implementations would miss to send a notification otherwise.
[...]

PR #440, opened 2018-11-15
use longer IMAP-idle timeout

r10s:
due to exponential backoff and dc_maybe_network(), we can process jobs at any time now, very short IMAP-timeouts are no longer needed (the 1 minute came from the need to process jobs in a very simple way); reset to 23 minutes as before imap/smtp-thrad-split

in addition to that, the UI can call dc_interrupt_imap_idle() at any time (to protect against  NAT Gateway timeout? not sure, but needed eg. in Android for any reason, cmp. https://www.isode.com/whitepapers/imap-idle.html)

### DIFF
--- a/src/dc_imap.c
+++ b/src/dc_imap.c
@@ -634,9 +634,9 @@ void dc_imap_idle(dc_imap_t* imap)
 		}
 
 		// most servers do not allow more than ~28 minutes; stay clearly below that.
-		// a good value is 23 minutes.  however, as we do all imap in the same thread,
-		// we use a shorter delay to let failed jobs run again from time to time.
-		#define IDLE_DELAY_SECONDS (1*60)
+		// a good value that is also used by other MUAs is 23 minutes.
+		// if needed, the ui can call dc_imap_interrupt_idle() to trigger a reconnect.
+		#define IDLE_DELAY_SECONDS (23*60)
 
 		r = mailstream_wait_idle(imap->etpan->imap_stream, IDLE_DELAY_SECONDS);
 		r2 = mailimap_idle_done(imap->etpan);


hpk42:
nice, looking forward to seeing this in action!

r10s:
> nice, looking forward to seeing this in action!

@hpk42 i think i will release a new android-dev version next week, have to fix the background issues before.
btw. have you experiences about how eg. python is responsive after some days in IDLE? would be very interesting. i tested this some time ago with the cmdline tool, however, not recently.

testbird:
The linked whitepaper says the connection needs to be refreshed before 30 minute server timeouts, **and** a NOOP is needed before 15 minute NAT routing timeouts.

Does deltachat support NOOPs? Possibly initiated by some servers?

r10s:
> Does deltachat support NOOPs? Possibly initiated by some servers?

libetpan supports it by `mailimap_noop()`, so, this won't be a problem as such.

imu while IMAP-IDLE is called, the socket waits for a response - how can we then send NOOP on the same socket? not sure how this can work (cc @hpk42)

however, if the ui calls `dc_interupt_idle()`, this already results in several IMAP commands, so i would assume an explicit NOOP command is not needed as there are several commands sent. in the RFC i do not see a concrete relation between IDLE and NOOP - file:///home/bpetersen/messy/books/imap-etc/RFC-3501-IMAP.html#section-6.1.2 vs. https://tools.ietf.org/html/rfc2177 - but i would be glad to have more information about this.

calling `dc_interrupt_idle()` from time to time may also protect against other things that may "hang" as it forces an interruption on the socket and keep the whole thread alive in some way. at least on andoid, this seems to work, for whatever reason, and avoids the thread to stay alive.

btw. i am also not sure if the "whitepaper" is correct, it's just a random link, there are also [comments that say NOOP is not needed in combination with IDLE](https://stackoverflow.com/questions/41433690/when-to-use-noop-command-to-keep-alive-imap-idle-connection).

testbird:
> if ui calls dc_interupt_idle(), this already results in several IMAP commands, so i would assume an explicit NOOP command is not needed as there are several commands sent.

Hm, yes the important thing is that some traffic occurs every 14 minutes.

The second link in https://github.com/deltachat/deltachat-core/issues/193#issuecomment-397884223 is good, it shows that at least Dovecot has an imap_idle_notify_interval setting to have the server send a "* OK still here" (444 bytes over SSL) to the client, which would be perfect if happening only about every 5 or 14 minutes to hold the NAT open. (A complete IDLE renewal from `dc_interrup_idle()` is at least 1,292 bytes over SSL, and the thunderbird stack will automatically renew the IDLE connnection if receiving the "* OK still there".)

So, the question is if deltachat can detect if the server initiates some activity in appropriate intervals <15 minutes, and otherwise do a manual most lightweight DONE, and restart IDLE every 14 minutes.

testbird:
> every 14 minutes

Hm, the comments to the answer https://stackoverflow.com/a/41433757 report 9 minutes for some mobile networks.

r10s:
well, i think there is not "correct way" or at least it is not easy to find.

seems as if we should just test things out and do them as correctly as we can as iterate from that.

testbird:
It's very simple, if deltachat is not reliable it will gets disqualified.

Buzzword iteration experiences or following denials (instead of solving known problems) does about the same, besides wasting time and efforts and creating annoyances.

With this thread we found some mobile networks need a keep alive interval < 10 minutes.
With others we found random exponential waiting used to delay resending ethernet frames for some amount of miliseconds will not work for delaying messages for minutes to days.

hpk42:
On Sun, Nov 18, 2018 at 07:50 -0800, testbird wrote:

> It's very simple, if deltachat is not reliable it will gets disqualified.

Not sure why you write this -- to suggest that we don't care about reliability? 
There is a lot of work on increasing reliability. 

> Buzzword iteration experiences or following denials (instead of solving known problems) does about the same, besides wasting time and efforts and creating annoyances.

Once again, please try to tone down. Iterating on algorithms, including timeout-constants,
is a perfectly valid approach to dealing with (flaky) networks and servers. 
 
> With this thread we found some mobile networks need a keep alive interval < 10 minutes.

It's one data point from a comment of a stack-exchange post. 
It seems that other MUAs use 23 minutes ... would be curious which ones btw.
Maybe the next try is 9 minutes or trying to be adaptive about idle-timeouts. 

> With others we found random exponential waiting used to delay resending ethernet frames for some amount of miliseconds will not work for delaying messages for minutes to days.

If exponential-backoff causes problems we can try something else.  
My understanding is that exponential backoff has more general applicability
despite ethernet-resending indeed often being the prime example. 
Re-trying less and less often after each failed attempt generally sounds
useful to avoid wasting network and cpu cycles.

Lastly, please consider that your repeated negative comments are not
helping as much as you might intend them to. It requires a lot
of time for maintainers to deal with github and support forum issues,
apart from actually developing code and designs and real-life testing ...
Please try to not waste it.

testbird:
I wrote the post because ideas of "no correct way" or "not easy to find" are a way that can make it hard for people to realize mistakes, and solutions, even if pointed to it, and I think stepping aside, or considering an outside voice, has more benefits  then hitting the wall.

Lets say there was just a glitch in managing the feedback and project professionally, consistently from raised issues to solutions, instead of from emotions to defense or arguments.


> other MUAs use 23 minutes

That's the idle timeout of K-9, also according to my link above. It may be increased to approach 30 minutes. That's different from the needed routing-keep-alives (9 minutes) that *may* even come from the server (dovecot: every 2 minutes).

> Iterating on algorithms, including timeout-constants, is perfectly valid

Sure, but a waste of time and effort if done in vain. If it can be anticipated that *randomized" (non-deterministic) sending delays in the first minutes can easily grow to long delays (a couple of failing very short (random) waits, followed by a long wait) will be very confusing and disappointing. Not even imagining what happens if the message order would get shuffled.

And it probably wont significantly reduce the battery and network usage either, as there was not much battery drain with a 1 minute idle timeout.

Traffic-wise it seems important that
* permanent errors from the smpt server will cancel the job (Looks as if message size and space errors are catched.)
* and interrupted transmissions can be continued. (I don't know if that's supported.)

Battey-wise it could be enough to 
* Do rather quick retries to process all jobs, with only up to max. 7 seconds delay before retrying. (Could be (non-randomized) exponential retries, but complicating is not necessary, the collision avoidance is already done on the network level.)
* Count the number of failures with a "circuit breaker" that interrupts the retries for longer periods (e.g. 6 up to 10 fails => 30 seconds,  up to 14 => 1 minute, up to 20 => 5 minutes, and 10 minutes afterwards)
* Let any new online status (network access) only trigger a retry (single one) if the circuit breaker has already waited for 1/3 of the current timeout ("half-open" circuit state).

See https://dzone.com/articles/understanding-retry-pattern-with-exponential-backover

testbird:
The algorithm in the last post is much better to **continuously** ensure that messages are sent as soon as possible, while still reducing the attempts with 3 minute timeouts when traveling weak flaky networks, and 10 minutes with good network strength (long lasting server/routing problems).

All the while any new network connection (wifi) can trigger dc_maybe_network() and cause the messages to be sent immediately.

It fulfills what users expect from a messager (quick sending and battery saving). Don't expect much qualified user feedback, unreliable and delayed messengers tend to just get uninstalled. Consider the bad effects of shipping something you know to do better are multiplied by the number of downloads and the effects multiplicators. Such basic things are really worth developer time.

r10s:
sending messages in strict order may cause problems if a given message cannot be sent for some reason. this would block all other messages then.

testbird:
May the smtp error code for permanent errors (larger error codes) allow to fail those jobs immediately?

Concerning non-permanent errors, messages may be put to the end of the queue until the first message in the queue is already marked as having failed with a non-permanent error. Meaning all messages have errors, and the original order is restored again, thus messages are sent in original order again once the (network) failure disappears.

testbird:
Much easier to implement: sort the messages in the queue by a second sorting criteria, the number of non-permanent failures.

testbird:
Then let each retry attempt loop over every level of non-permanent failures. Starting with the messages with the most failures (the oldest ones), and then trying later messages.

testbird:
The number of non-permanent failures for a message (i.e. not an address error etc.) may also provide better feedback for the user, e.g. if shown with the clock symbol in messages waiting to be sent.

Allowing users to go to a wifi login page, to the window, leave the house, ... or cancel the message.

PR #442, opened 2018-11-16
refine dc_configure() doc

### LINKED ISSUE #246: keep config parameters in sync if dc_configure() fails
a typical configuration process is

```
dc_set_config(context, "addr", "alice@example.org"); // use some real test credentials here
dc_set_config(context, "mail_pw", "***");
dc_configure(context);
```

what `dc_configure()` does is the following:
- reads `addr`, `mail_pw` etc. from the database
- fills gaps in the setup (missing ports, user names, etc.) using eg. autoconfigure, autodiscover, defaults and testing
- if a working configuration was found, this is saved completely as WORKING, so that subsequent connects can use the correct parameters right away

this is fine on the first configuration, however, if a subsequent configuration _fails_, eg. `mail_server` may not match the WORKING one - this is no problem internally, however, typically UI implementation may want to show a dialog with the current parameters and there is no way to get them after dc_configure() failed one time (again the lib as such is still working)

solutions for this would be:
1. tell the UI just not to _read_ the `addr` etc.
2. pass all parameters directly to dc_configure()
3. reset  `addr` etc. to WORKING if dc_configure() fails and document this

=> 1. would have the disadvantage that there is no way to get the configuration from the lib, the UI would need to track the values on its own; 2. seems to be the clearest approach, however, it will make bindings etc. more complicated. also there will be functions as get_whole_config() needed which would make the whole interface more complex.

@ralphtheninja so, i would fo for 3. which does not require large changes and keeps things simple and the get/set straight-forward from the view of the UI.

r10s:
closes #246

### DIFF
--- a/src/dc_configure.c
+++ b/src/dc_configure.c
@@ -722,34 +722,49 @@ cleanup:
 
 
 /**
- * Configure and connect a context.
- * For this, the function creates a job that is executed in the IMAP-thread then;
+ * Configure a context.
+ * For this purpose, the function creates a job
+ * that is executed in the IMAP-thread then;
  * this requires to call dc_perform_imap_jobs() regularly.
  *
- * - Before you call this function, you should set at least `addr` and `mail_pw`
- *   using dc_set_config().
+ * - Before you call this function,
+ *   you must set at least `addr` and `mail_pw` using dc_set_config().
  *
- * - Use `mail_user` to use a different user name than `addr` and `send_pw` to
- *   use a different password for the SMTP server (otherwise `mail_pw` is used)
+ * - Use `mail_user` to use a different user name than `addr`
+ *   and `send_pw` to use a different password for the SMTP server.
  *
- *     - If _no_ more options are specified, the function tries to get these
- *       **using autoconfigure/autodiscover**.
+ *     - If _no_ more options are specified,
+ *       the function **uses autoconfigure/autodiscover**
+ *       to get the full configuration from well-known URLs.
  *
- *     - If _more_ options as `mail_server`, `mail_port`, `send_server`, `send_port`, `send_user` or `server_flags`
- *       are specified, **autoconfigure/autodiscover is skipped**.
+ *     - If _more_ options as `mail_server`, `mail_port`, `send_server`,
+ *       `send_port`, `send_user` or `server_flags` are specified,
+ *       **autoconfigure/autodiscover is skipped**.
  *
- * - While dc_configure() returns immediately, the started configuration-job may take a while,
- *   you can stop it using dc_stop_ongoing_process().
+ * While dc_configure() returns immediately,
+ * the started configuration-job may take a while.
+ * You can stop it using dc_stop_ongoing_process().
  *
- * - The function sends out a number of #DC_EVENT_CONFIGURE_PROGRESS events that may be used to create
- *   a progress bar or stuff like that.
+ * During configuration, #DC_EVENT_CONFIGURE_PROGRESS events are emmited;
+ * they may be used to create a progress bar.
+ *
+ * On a successfull configuration,
+ * the core makes a copy of the parameters mentioned above:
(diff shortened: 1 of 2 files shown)


src/dc_configure.c
```diff
@@ -722,34 +722,49 @@ void dc_job_do_DC_JOB_CONFIGURE_IMAP(dc_context_t* context, dc_job_t* job)
...
+ * after a configure-attempts has failed.
+ * Otherwise the parameters may not reflect the current configuation.
  *
  * @memberof dc_context_t
  * @param context The context object as created by dc_context_new().
  * @return None.
  *
- * There is no need to call this every program start, the result is saved in the
- * database and you can use the connection directly:
+ * There is no need to call dc_configure() every program start,
```
ralphtheninja: on every

src/dc_configure.c
```diff
@@ -722,34 +722,49 @@ void dc_job_do_DC_JOB_CONFIGURE_IMAP(dc_context_t* context, dc_job_t* job)
...
- *       are specified, **autoconfigure/autodiscover is skipped**.
+ *     - If _more_ options as `mail_server`, `mail_port`, `send_server`,
+ *       `send_port`, `send_user` or `server_flags` are specified,
+ *       **autoconfigure/autodiscover is skipped**.
  *
- * - While dc_configure() returns immediately, the started configuration-job may take a while,
- *   you can stop it using dc_stop_ongoing_process().
+ * While dc_configure() returns immediately,
+ * the started configuration-job may take a while.
+ * You can stop it using dc_stop_ongoing_process().
```
hpk42: What does a second call to dc_configure() do? Does it simply not execute or does it put a second job to imap?  
why does dc_configure() not call dc_stop_ongoing_process() impilicitely?

hpk42: and if the context/db is already configured then dc_configure is a no-op, right? this should be mentioned.

r10s: > What does a second call to dc_configure() do? Does it simply not execute or does it put a second job to imap?

it puts a second job to imap. 
does not make sense, however.

> why does dc_configure() not call dc_stop_ongoing_process() impilicitely?

it does. an explicit call is only needed to stop an ongoing configuration job before it is finished.

> and if the context/db is already configured then dc_configure is a no-op, right? this should be mentioned.

no - it just configures again. this is useful eg. if the user wants to change some settings, password, options, whatever.

r10s: refined the doc and avoid enquing a second config-job while the running is not yet finished.

src/dc_configure.c
```diff
@@ -786,6 +812,19 @@ int dc_is_configured(const dc_context_t* context)
 }
 
 
+/*
+ * Check if there is an ongoing process.
+ */
+int dc_has_ongoing(dc_context_t* context)
```
hpk42: why "process"?  Is this True for both configuration and verification or so? 
IMHO It's better if it's a specific API about configure like "dc_configure_is_ongoing"
and "dc_verification_is_ongoing" (if needed at all).

r10s: it's true for configure, import and export. these "processes" are mutual exclusive.

r10s: btw. the whole ongoing-function-group is mainly for allowing the interruption of the configure-/import-/export-/key-transfer-processes (interrupting is just done by checking a flag from time to time, typically together with emitted process events). 
i liked to have _one_ function that can interrupt such things, sth. as CTRL-C :)

the check to _not_ start a process if another is still running is more a side effect - if we skip it completely, the jobs would be just enqueued, however, i do not think the user of the lib would do this intentionally, so a check makes some sense - and is rather cheap as the ongoing-stuff is already "allocated" to allow interruption.

hpk42:
ok, don't want to hold this up further.  still not sure, i get 100% of how the API behaves but things can always be improved in other PRs ;)

PR #446, opened 2018-11-21
 tweak get_backoff_time_offset()

r10s:
tweaking/fixing  get_backoff_time_offset():

- moved the `- 1` to the exponent (probably a typo, does not make really sense outside the bracket imho) so that the first retry is within a minute (2^0  = 1)
- to keep the max. timespan, increased the number of retries by one
- move the MULTIPLY in front of the randomization; before, every first-retry was exactly either 0 or 60 seconds

note, however, that each failed try is repeated _at once_, so that closed sockets and such things are catched immediately. 

moreover, _if_ the ui knows network becomes available again, the ui may bypass the calculated times by using dc_maybe_network(). on desktop, afaik, this function is currently never called, so that sending will be always delayed by these timespans.

### DIFF
--- a/src/dc_job.c
+++ b/src/dc_job.c
@@ -413,13 +413,13 @@ static void dc_suspend_smtp_thread(dc_context_t* context, int suspend)
 static time_t get_backoff_time_offset(int c_tries)
 {
 	#define MULTIPLY 60
-	#define JOB_RETRIES 16 // results in ~3 weeks for the last backoff timespan
+	#define JOB_RETRIES 17 // results in ~3 weeks for the last backoff timespan
 
-	int N = (int)pow((double)2, c_tries) - 1;
+	time_t N = (time_t)pow((double)2, c_tries - 1);
 
-	int r = rand() % (N+1);
+	N = N * MULTIPLY;
 
-	time_t seconds = r * MULTIPLY;
+	time_t seconds = rand() % (N+1);
 
 	if (seconds<1) {
 		seconds = 1;


hpk42:
makes sense i think.

testbird:
The former -1 may have originally been intended to allow the first attempt (retry 0) to be executed without any delay (2^0 - 1=0), but it seems you may have introduced or are using DC_AT_ONCE for that.

Sending reliably as soon as possible, may also be rather simple without larger changes to the current code structure:

* remove randomization line (let delays stay deterministic and keep order of delayed jobs)
* have a tabular or cased get_backoff_time_offset() 
```
case c_tries
    up to  5    =>  5 seconds
    up to 10    => 30 
    up to 14    => 60 
    up to 20    => 300 (5 minutes)
    above 20    => 600 (10 minutes for retries beyond the first half hour)
```
* increase JOB_RETRIES > 720 (about 5 days)

* And too many and fast auto-retries on flaky network connections may be limited by having that trigger call the the same get_backoff_time_offset() with the retry count of the next job, and only do a retry if the job is already due in less than 2/3 of the returned up offset.

csb0730:
Hi everyone,
months ago I sent an email to r10s to improve resending behaviour of v0.20.0 which has been unacceptable (Three retries, 1min fix interval and thats it!).

My new idea was to increase retry counter only at errors where at minimum two consecutive retries are under "network available" conditions. See this illustration:

```
--------------------------------------------------------------

retry limit     5                                A--
                4....................     ----  /
                3                        /    \/
                2                       /
                1    ----          --  /
retry counter	0---/    \--------/  \/
--------------------------------------------------------------
retry              S+    + + +   ++  +++++    ++++EEEEEEE
--------------------------------------------------------------
Network state	11111000010101000110011111000011111111110

--------------------------------------------------------------

+ = retry sending msg
0 = Network State NOK
1 = Network State Ok
A = Abort retry
S = Start of send (first try)
E = msg in error, no further retry
```
With this code I build an own apk which I have in service for months now. IMHO with good success :-)
This code based on the old `is_online()` function.

See here
- sources (line 625-648) https://github.com/csb0730/deltachat-core/blob/master/src/dc_job.c
- apk https://github.com/csb0730/deltachat-core/releases/

While I adaped the code team worked on that issue too but took another approach: took is_online() away and choose the current exponential approach. I think meanwhile experience with this code is available too.

Everyone is invited to test, but attention: export backup and import is required to keep messages and settings (or new installation).

Maybe my own approach can help in the future?

adbenitez:
Have to say I am using 0.98.2 and still find DC struggled at sending some time it is immediately, but when a network error occurs, I found myself closing and reopening the app to force the send to happen immediately, otherwise it can take several seconds, once on some 0.9X version took 13min!! 
Maybe there are nothing to do with this and it is just a socked timeout set this way by design...

testbird:
Hi, good points @csb0730 and @adbenitez, I haven't looked back at deltachat, but maybe this year is better for implementing solutions rather and worse for closing uncounted valid issue reports.

As written several times, and you of course found out as well, the backoff time intervalls in dev version are getting way too long too quickly for a chat app. Actually a small thing simple to acknowledge.

The "low pass filter" idea to only count "two consecutive online reties" is a really nice idea, but a drawback may be that it is still likely one can get a "EEEE" (error) state, requing user unnecessary intervention (resending).

That's why think it would be better (and possibly also easier) to just have the backoff time intervall increase in some some reasonable steps, and finally continue retrying (after the _non-permanent_ smtp errors) with an interval that is long enough to prevent draining the battery _too much_ (i.e. only creating system activity that reduces a phone's 100h stand-by to to a 99h stand-by, see https://github.com/deltachat/deltachat-core/pull/446#issuecomment-441385285). Compared to the low pass filter, that should also have an advantage when dealing with connection jitter, by completely avoiding too many retries when the network is flaky.

hpk42:
Hi Christian, 

On Mon, Jan 21, 2019 at 15:51 -0800, Christian Schneider wrote:
> Hi everyone,
> months ago I sent an email to r10s to improve resending behaviour of v0.20.0 which has been unacceptable (Three retries, 1min fix interval and thats it!).
> 
> My new idea was to increase retry counter only at errors where at minimum two consecutive retries are under "network available" conditions. See this illustration:
> 
> ```
> --------------------------------------------------------------
> 
> retry limit     5                                A--
>                 4....................     ----  /
>                 3                        /    \/
>                 2                       /
>                 1    ----          --  /
> retry counter	0---/    \--------/  \/
> --------------------------------------------------------------
> retry              S+    + + +   ++  +++++    ++++EEEEEEE
> --------------------------------------------------------------
> Network state	11111000010101000110011111000011111111110
> 
> --------------------------------------------------------------
> 
> + = retry sending msg
> 0 = Network State NOK
> 1 = Network State Ok
> A = Abort retry
> S = Start of send (first try)
> E = msg in error, no further retry
> ```

very interesting graphic, thanks!  While we are not going to change the logic
for the pending releases the retry-sending logic needs scrutiny.  

FWIW I think we need better feedback of what's happening in terms of
smtp/imap retries/reconnects etc. Currently it's all buried in the
log file and hard to interpret, even for devs.  In another project,
that deals with network replication, we introduced a "server status":
https://m.devpi.net/
If you click on "server status" you'll see the state of network replication.

Something like this would be useful to have -- an icon you can enable
in advanced settings (green/maybe-orange/red) and on click you'd see the history of tries, 
number of pending messages -- last connects and outcomes via smtp and imap etc. 
this would give advanced users more precise information to feedback
from their particular situation.  Green would mean -- connections are fine,
no pending outgoing messages. As Asiel, i also had situations where i force-quitted
the 0.99 app and felt that afterwards things started to flow better.  It's unclear
what exactly was the problem, though.  So i think we need better per-device/situation
debuggability to improve things in an informed manner.  Makes sense? 

> With this code I build an own apk which I have in service for months now. IMHO with good success :-)
> This code based on the old `is_online()` function.
> 
> See here
> - sources (line 625-648) https://github.com/csb0730/deltachat-core/blob/master/src/dc_job.c
> - apk https://github.com/csb0730/deltachat-core/releases/

could you adapt your logic to not poll status via dc_is_online()? 
Right now, when there is a "network available" event Android triggers 
resending logic but there is no way to "poll" network-status via the core anymore. 

Again, thanks for helping/trying other approaches out, much appreciated!

holger

testbird:
Err, I may add: My kudos and contratulations to getting the new dev version ready for release so quickly.

My "complains" really only mean that I do think that problems that got identified and reported could be handled better. It's a pity seeing @adbeniz wasting time and stumbling over one known and closed issue after the other.

testbird:
> i think we need better per-device/situation debuggability to improve things in an informed manner. Makes sense?

I would only direct resources into such instrumentation, if it would still needed after the known problems are fixed.

As @r10s explained, the current code has a problem with using wrong start values in it's exponential randomization calculation. As this calculation is not necessary as it is for collision resolution in shared bandwidth, I table should do.

csb0730:
@hpk42 

> could you adapt your logic to not poll status via dc_is_online()? Right now, when there is a "network available" event Android triggers resending logic but there is no way to "poll" network-status via the core anymore.

Hi Holger,
the intension behind my approach are these assumptions:
1. "is_online()" is not 100% reliable (this is proven by behaviour of v0.20.0 with it's first simple approach)
2. when "is_online()" fails there is the assumption, that try before too has been done with bad network conditions so retry counter needs to decreased again. 
3. only real server problems should increase retry counter, not unsecure or unknown network conditions! This leads to the approach to say: If I have two consecutive errors under the condition network is ok at least 1st retry of these two has to increase retry counter!
4. If there is really a problem with server or mail size (for example) it makes sense to reach as soon as possible a decision point to abort retries!
Therefore I increased retry counter only from 3 to 5, not more. Any other decision shoud be done by user (But this is not done now).

With old v0.20.0 approach I saw very often msgs not send ([!] icon). Since I introduced my approach I never saw a not send message. Even under bad conditions: weak net and/or pictures to send with bad net!

Your question to adapt this approach with "network available" event?
Essential for this logic is, that you need to know network state for every retry. If we can get a this information somehow it's possible. By the way, I think any reasonably approach can only based on top of this information!

A next essential thing is (as You mentioned) getting sufficient information for developpers about issues currently in focus.
So, the easiest and most transparent way is using the DC's internal log-ring-buffer for that !

With that, user has the opportunity to check immediately what happens without adp pain at every time something strange happens. It can be copied by current clipboard funtion to a file or whereever and save it for examinations.
In my apk I extended the buffer, so that at least 2h under high usage conditions are available again.

My experience with internal device log and adp is that important informations are overwritten until I got the chance to extract it (at evening) from the device!

adbenitez:
I also think the logs are overridden too quickly.
Also I don't know if my network was in its best times but I was not able to see  disconnections/reconnect logs.

adbenitez:
Forget everything I have said!!! Stop wasting time and resources in logs development and meet MatLog Libre!!!

csb0730:
btw, log-ring-buffer is available, not to develop ;-)

But what is MatLogLibre ?

adbenitez:
Christian Schneider:
Yeah, I was a bit like joking :)
What I meant is that devs don't need to waste time with the log ring view  in Delta Chat settings to satisfy our needs, when we can just use Matlog Libre which is an fdroid app to  see app logs, with features like automatically removing sensitive information like URLs, phone numbers and email addresses, you can easily save to a file and even compress it, you can set the ring size to whatever you want, has filters, etc.
So why are you wasting your time with the basic dc's build-in log view???

adbenitez:
Also, the DC's log view does anything else but show static logs, you have to go back and forth over and over so it is updated which is really annoying :-/

testbird:
> 4. If there is really a problem with server or mail size (for example) it makes sense to reach as soon as possible a decision point to abort retries!

As stated above connection errors are already handled separately, and I think @r10s also had  commits to catch size errors separately.

Consequently I think our discussion and the issue https://github.com/deltachat/deltachat-android/issues/602 is just about letting deltachat handle "temporary smtp sending errrors" properly.
* keep retrying with reasonably short intervals first, for quick delivery.
* increase intervals up to a aceptable delay/battery ratio
* don't waste retries in flaky conditions

csb0730:
@adbenitez
I like DC's internal log-ring-buffer because it's easy to use (developper and user) and available immediately for every user!
I don't like too many apps at my phone! Especially not for standard purposes which can be handled without them too. In this example here :-)

adbenitez:
Christian Schneider:
For developers Matlog Libre is far superior to the DC's internal log-ring-buffer, users may be more interested in just sending the bug report or log not to read it :P so the current one is enough.
 To me it is more easy to switch between DC and Matlog than going back and forth from a chat room to settings/advanced/view-log not to say that DC logs view doesn't updates automatically do you must keep going back and reopening  it to see if there are some new logs, I suggested it because you seem not happy with the DC log-ring view, so I was giving you an option that let you customize the ring size and a lot of cool stuff, if you are happy with it, then sorry, just forget I mentioned it,
This is really off topic so pls mail me further replies instead of replying here :)

PR #456, opened 2018-11-27
Less brittle meson install, securer downloads

flub:
- This makes installing meson a little less brittle.  It's still not
  ideal I guess.
- Push sha256 checksums on all curl downloads, at least we'll all be
  running the same malware this way.
- Only copy the deltachat wheels to the wheelhouse, that avoids having
  to clean up the unintersting wheels.

### DIFF
--- a/python/wheelbuilder/Dockerfile
+++ b/python/wheelbuilder/Dockerfile
@@ -7,14 +7,19 @@ ENV PKG_CONFIG_PATH /usr/local/lib64/pkgconfig:/usr/local/lib/pkgconfig
 
 # Install meson and ninja
 ENV NINJA_VERSION v1.8.2
+ENV NINJA_SHA256 d2fea9ff33b3ef353161ed906f260d565ca55b8ca0568fa07b1d2cab90a84a07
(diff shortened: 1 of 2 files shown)


hpk42:
thanks a lot, @flub! Now i guess it'd be interesting to integrate this building into Jenkins and/or Travis.

flub:
That's an interesting question, do you think it is worth setting this up to automatically produce wheels?  If so where should they be published?  And what to do about tagged releases and pypi?

This should probably be discussed on the mailing list instead of here though ;)

hpk42:
On Wed, Nov 28, 2018 at 01:56 -0800, Floris Bruynooghe wrote:
> That's an interesting question, do you think it is worth setting this up to automatically produce wheels?  If so where should they be published?  And what to do about tagged releases and pypi?

I thought about a https://m.devpi.net/deltachat/dev index.  One thing to get right
then is to get automatic versioning correct -- i guess setuptools_scm is the latest
way to do this? 

> This should probably be discussed on the mailing list instead of here though ;)

maybe.  It's a bit of a special topic, so doing it in an issue is fine as well.

Issue #444, opened 2018-11-19
send own avatar to recipients


r10s:
the api already allows to save a profile image ("avatar") of the user in the same way as the user's display-name or the user's status, see [dc_set_config(context, "selfavatar", ...)](https://c.delta.chat/classdc__context__t.html#aff3b894f6cfca46cab5248fdffdf083d)

while locally reading and writing the avatar is implemented (and is already used by android-dev), it's an open question how the avatar data can be **sent to the recipients:**

- one possible way to do this is defined in the spec at https://github.com/deltachat/spec/blob/master/spec.md#set-profile-image - this would add the avatar as a normal attachment, same as for the group-images. i think this method makes some sense.

- an alternative may be the X-Face header, http://quimby.gnus.org/circus/face/, which however, is very limited in dimension and image size otoh it is displayed also by other MUAs. but this advantage would probably also disappear if we start to move the header to the encrypted part.

another open question is **how often** the avatar data is sent to the recipients:

- maybe just on changes, only once per recipient
- for this purpose,  a flag to track the avatar-sent-state would be required

r10s:
> Could a MUA not still decrypt the header and display that avatar?

they _could_, however, they just _don't_ as far as i know.
i think there a few MUAs using `X-Face` at all, and i don't think there is a single one that reads  `X-Face` from the encrypted part (using memoryhole).

PR #461, opened 2018-11-30
enhance profile-image api

r10s:
- `dc_chat_get_profile_image()` returns the contact image for normal chats
- `dc_contact_get_profile_image()` added; currently the own profile image is returned for the SELF-chat (can be set using `dc_set_config(context, "selfavatar", image)`.
in the future, it is planned to return also the contact images received from remote users here, see #444


ralphtheninja:
>     * `dc_chat_get_profile_image()` returns the contact image for normal chats

Nice! This will make it purdier.

PR #466, opened 2018-12-05
get chat and contact colors

r10s:
this pr adds two functions that return "fine" colors for chats and contacts.
the colors can be used for drawing avatars or the names above messages in group chats.

### DIFF
--- a/src/dc_chat.c
+++ b/src/dc_chat.c
@@ -260,6 +260,46 @@ cleanup:
 
 
 /**
+ * Get a color for the chat.
+ * For 1:1 chats, the color is calculated from the contact's email address.
+ * Otherwise, the chat name is used.
+ * The color can be used for an fallback avatar with white initials
+ * as well as for headlines in bubbles of group chats.
+ *
+ * @memberof dc_chat_t
+ * @param chat The chat object.
+ * @return Color as 0x00rrggbb with rr=red, gg=green, bb=blue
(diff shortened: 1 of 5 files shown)


src/dc_chat.c
```diff
@@ -259,6 +259,46 @@ char* dc_chat_get_profile_image(const dc_chat_t* chat)
...
+ * For 1:1 chats, the color is calculated from the contact's email address.
+ * Otherwise, the chat name is used.
+ * The color can be used for an fallback avatar with white initials
+ * as well as for headlines in bubbles of group chats.
+ *
+ * @memberof dc_chat_t
+ * @param chat The chat object.
+ * @return Color as 0x00rrggbb with rr=red, gg=green, bb=blue
+ *     each in the range 0-255.
+ */
```
ralphtheninja: You mean from `0` to `ff`?

violoncelloch:
And as background color for the message bubbles?

src/dc_chat.c
r10s: jupp :)
do you think we should better write it in hex?

r10s:
> And as background color for the message bubbles?

might be possible, however, the plan is to use white/light backgrounds in the bubbles.

src/dc_chat.c
ralphtheninja: Nah, it's probably fine :)

Issue #422, opened 2018-11-06
support a fallback of using the DeltaChat folder similar to the past


r10s:
the single-folder approach may have the potential to annoy exiting users that are used to the `DeltaChat` folder.

we should add a way to the core library that support the workflow of moving messages from the INBOX to another folder, so that in a shared-usage, the INBOX stays empty.

we could do this by using a 3rd thread that watches the INBOX (in addition to the [configured single-folder](https://github.com/deltachat/deltachat-core/pull/407)) and moves all messages in the INBOX with the `Chat-Version:` header to the folder configured as the single-folder.
i would not do much more as this would make things much more complicated (eg. folder creation (folders may be hidden, not hidden, subfolders of the INBOX and so so , this would require more maintainance) automatic upgrade and so on)

@angelo-fuchs on the android-ui i would suggest a checkmark in the advanced account settings, together with an option to type in the folder.

cc @hpk42

testbird:
Thank you for re-considering to do some basic message sorting in the -core. Good to see you found the time for some real reasoning about this. As deltachat knows most about the message status, I think it is a very well suited place to do the sorting intelligently, providing a good overall user exprience.

I think it was @t-oster who first found out and formulated that the INBOX would always be the main folder to watch (https://github.com/deltachat/deltachat-core/issues/117#issuecomment-395687590). It allows to detect incoming (short chat) email messages, and replies from classic MUAs, and to move messages into a separate folder after chat messages have been displayed or email messages were answered in a chat (also allows to (intermittently) continue a chat from a classic email client, even with concurrently running multiple chat and email clients). While the "AltMarkMove" scheme (https://support.delta.chat/t/wiki-imap-strategy-to-live-within-the-email-ecosystem/123) also reduces the distubancees of email client listings (all chats immediately flaged "seen") and only requires a *single* folder watch.

If we leave out all the originally SIEVE ideas in the wiki, and the detection of "email-approved" (known contacts), the basic things do stay plain simple and manageable. But as the long list of isolated points may not have made it obvious how the concurrent syncing could work, and unfortunately nobody asked, I added the following explanation to the "basic approaches" section:

[...]

Issue #408, opened 2018-10-24
check Message-ID before downloading messages from IMAP-server


r10s:
due to #370 of https://github.com/deltachat/deltachat-core/milestone/2 , we save about half of the sending bandwidth by uploading a copy using "bcc" instead of a separater IMAP-upload.

while this is working, currently, the saved bandwidth is used for re-downloading the message as soon as IMAP-IDLE wakes up to notice us about the message we've just sent.
the message is not added to the database, however, it is downloaded before aborted.

the very most of this additional bandwidth can be saved by checking the Message-ID before downloading the body. IMAP has some methods for this i cannot tell from mind :)

nb: thinking it over, this additional download was also done in the past - eg. the IMAP-upload is done to the SENT-folder and this folder - as well as all others - are scanned about every 30 minutes.

so already now, we save 1/3 compared to all versions up to now (0.20). with this optimization we will save 2/3 :)

r10s:
@bpiddubnyi great that you want to have a look at this :)

not directly related, but when working on the loop that downloads from the IMAP folder: 

there may be a bug in it, a race condition or whatever. in the past, it happened some times that old messages get downloaded randomly, see https://github.com/deltachat/deltachat-core/issues/156. i always assumed sth. with the stored "last-id", however, i did not found anything (hard to find errors on code written by oneself :) 
however, might also be that the bug is removed by the code removal/changing of https://github.com/deltachat/deltachat-core/milestone/2

testbird:
Another related possibility to skip downloading the message body and to reduce the mobile traffic are subject-only messages #437 and attachments #436 .

PR #473, opened 2018-12-13
[wip] move chat messages away

r10s:
this pr is about to re-implement the move between the INBOX and the DeltaChat folder, see #422

- [x] add a second thread (MVBOX-thread) that watches the DeltaChat folder
- [x] expose the api to set different folders for different threads 
- [x] make the old INBOX-thread folder configurable
- [x] find out the correct MVBOX- resp. DeltaChat-folder during configure - depending on the server, eg. only one of INBOX/DeltaChat or DeltaChat is allowed ... IMAP :/
- [x] make sure, the MVBOX-folder is also set for already configured accounts (upgrades from f-droid)
- [x] use the MVBOX-folder
- [x] make the use of MVBOX optional to allow a pure-dedicated account as just now
- [x] make sure, calling dc_perform_mvbox*() is fine if the MVBOX is disabled (might be hard for the ui to not call these functions, which, however, would be the best)
- [x] markseen() etc. is always called as a job in the INBOX thread - make sure this works also with messages in the DeltaChat folder (should be  - a SELECT is needed for various reasons anyway and the folder as such is known - be we should check this carefully)
- [ ] add functionality to move messages from INBOX to MVBOX (see #475 for a discussion about the timing)
- [ ] _carefully_ decide on the criterions about when to move

we should also consider implementing #408 (check message-id before downloading) as moving between two watched folders results in threads waking up and some additional traffic.

### DIFF
--- a/cmdline/main.c
+++ b/cmdline/main.c
@@ -126,21 +126,38 @@ static uintptr_t receive_event(dc_context_t* context, int event, uintptr_t data1
  ******************************************************************************/
 
 
-static pthread_t imap_thread = 0;
+static pthread_t inbox_thread = 0;
 static int       run_threads = 0;
-static void* imap_thread_entry_point (void* entry_arg)
+static void* inbox_thread_entry_point (void* entry_arg)
 {
 	dc_context_t* context = (dc_context_t*)entry_arg;
 
 	while (run_threads) {
-		// jobs(), fetch() and idle()
-		// MUST be called from the same single thread and MUST be called sequentially.
+		// jobs(), fetch() and idle() MUST be called from the same single thread
+		// and MUST be called sequentially.
 		dc_perform_imap_jobs(context);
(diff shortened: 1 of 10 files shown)


testbird:
Thank you, I'm really grateful.

I also agree a second idle watch is also ok to start with ("avoid too early optimizations").
Only found the MVBOX naming unecessarily confusing, when things could just be referred to as the INBOX- and CHAT- (folder) threads.

testbird:
> carefully decide on the criterions about when to move

What have you planned for the basic approach?

If moving is enabled, moving jobs would be scheduled based on INBOX-folder arrivals?
CHAT-folder "arrivals" can cancel moving jobs, deletions can be replicated localy?

hpk42:
k, so let's merge this already without the actual move-logic so we can base further tests on it.

PR #504, opened 2019-01-07
allow broadcasting own messages without marking them as being forwarded

r10s:
this pr stops marking messages as being forwarded if they are sent by the user himself.

this way, there is a simple "hack" to broadcast messages: 

- send a normal message to the first chat
- forward the message to all other chats that should also get the message

for the recipients, all messages appear in the same way and are not marked as "being forwarded".

(the "forward mark" is applied only to messages sent by other people then)

### DIFF
--- a/src/dc_chat.c
+++ b/src/dc_chat.c
@@ -2600,7 +2600,13 @@ void dc_forward_msgs(dc_context_t* context, const uint32_t* msg_ids, int msg_cnt
 				goto cleanup;
 			}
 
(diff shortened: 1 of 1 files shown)


r10s:
i think the failed check is related to the environment, not to the code changes.

excerpt from https://travis-ci.org/deltachat/deltachat-core/jobs/476349a322 :
```
E: Unable to locate package g++-7
E: Couldn't find any package by regex 'g++-7'
apt-get.diagnostics
apt-get install failed
```

Issue #261, opened 2018-09-05
Backup import not working properly - extensive memory usage


csb0730:
This is the continuitation of last comments to "Limit retained history" #244.
While discussing that this issue here appeared.

csb0730:
As I can estimate the memory consumption while importing backup, see this consideration:
- starting point: backup db size 1GB
- while importing db, this is copied so we come to a summary memory consumption of
  ==> 2GB (messenger.db + backup.db)
- now files are copied to directory (95% of db space needed to copy files out of db into blob dir)
  ==> 2GB + 0.95GB = 2.95GB in summary (messenger.db + backup.db + exported file attachments).
- now db wants to execute vacuum:
  ==> 2.95GB already committed and now we need another ~1GB for doing vacuum (!):
  ===> 3.95GB reached !
- If all processed runs ok we should see a messenger.db of around 100MB and a summary of memory consumption of ~1GB or a litle more.

At my device I see after importing done a memory consumption of 1.93GB and a messenger.db of 1GB !

Conclusion:
In spite of free memory of 2.6GB at device it's not possible to import a backup db of 1GB size!

Can anyone confirm this behaviour and memory consumption?
What's the best approach to handle this process better?

csb0730:
At PC, pulled messenger.db (adb) can be vauum'd (but not with the pure command directly!)
First I need to launch "end transaction" then "vacuum" !
That works and shrinks db from 1GB to 138MB !

r10s:
> 09-06 00:22:54.062 E/DeltaChat(17171): Cannot excecute "VACUUM;". SQLite says: disk I/O error

i think, this is the important line.
VACUUM needs at least as many free space as the size of the database. obviously even a bit more.

i think, the solution is quite clear and described already here, https://github.com/deltachat/deltachat-core/issues/244#issuecomment-417848121

r10s:
> Regarding solution: It could be a way too, to use a separate database file for dealing with the attachments. Maybe easier.

easier to program, maybe, but more difficult for the user who has to copy two files.
i think putting both into the same file is fine.

another idea that just came into my mind, is to put the original sqlite-database as a blob into the database that already contains all the images as blobs. might be very simple and straigt-forward.

csb0730:
First one word to the following line:
`09-06 00:22:53.971 E/DeltaChat(17171): Cannot excecute "END TRANSACTION;". SQLite says: cannot commit - no transaction is active`
[...]

PR #522, opened 2019-01-13
cleanup before export

r10s:
do housekeeping before exporting to avoid exporting files that are no longer needed.
moreover, a sqlite-vacuum is done, just in case that it has failed on import and there is now more memory available.

mitigates #261, as the exported file will by smaller. however, importing still needs up to twice of the size of the database to import.

### DIFF
--- a/src/dc_imex.c
+++ b/src/dc_imex.c
@@ -859,6 +859,12 @@ static int export_backup(dc_context_t* context, const char* dir)
 		}
 	}
 
+	/* delete unreferenced files before export */
+	dc_housekeeping(context);
+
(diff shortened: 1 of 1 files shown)


csb0730:
> mitigates #261, as the exported file will by smaller. however, importing still needs up to twice of the size of the database to import.

Remark: Import needs quite up to _four_ times of the size of db over all !
See #261 comment from 09/06/2018.

r10s:
>  Import needs quite up to four times of the size of db over all !

i did some calculations at https://github.com/deltachat/deltachat-core/issues/261#issuecomment-453991575

Issue #156, opened 2018-04-30
Old mails are downloaded sometimes


r10s:
Sometimes, Delta Chat starts to download old messages, i could not figure out the reason for this yet, seems to be an issue with the UID logic in the IMAP processing, https://github.com/deltachat/deltachat-core/blob/master/src/dc_imap.c

testbird:
How can this downloading be noticed?

(In case that delta chat got newly installed, it could actually be desirable to download the messages from the server. Then only the setup message could be enough to (re)set up a (new) device, no need to transfer backup anymore, then.)

r10s:
> How can this downloading be noticed?
well, there are old mails in the chats (older that the time Delta Chat was installed)

testbird:
> old mails in the chats (older that the time Delta Chat was installed)

Hm, somehow I would expect that from a decent email client: To show all the messages that are on the server. (Also needed e.g. to delete messages.)
Of course Email-chat clients may default to show only chat messages.

r10s:
"It's not a bug, it's a feature" :)

testbird:
> "It's not a bug, it's a feature" :)

Yeah, was the server import introduced, or rather solved :) to some degree recently?
I remember @Ampli-fier providing instructions that I did not manage to try out, before somehing was changed. https://github.com/deltachat/deltachat-android/issues/210

testbird:
In case you plan to further disable all downloading, to avoid filling up deltachat with huge archives:

Maybe instead just limit the inital download to the last 20 messages (even already seen) in INBOX, to not fill-up DC, and still have some current messages to be found by new users when they start exploring the app.

sketet:
hello, I really would like to show old emails - that I used as chat.
if I can't view them I can't find the reason to use this messenger.

in the faq there is:

> Which messages do appear in Delta Chat?
> 
> Delta Chat automatically shows …
> 
>     Messages from contacts in your address book
>     Messages from contacts contacted by you
>     Replies to messages send by you
> 
> Other messages do not appear automatically. You can see the other messages in the main menu at Contact requests and, if desired, start a chat from there.


I've the contacts in the addressbook, but no one email is showed.

testbird:
I think we all agree, and need this. But it needs work to get implemented. @Ampli-fier tried to trigger downloading manually, and it seemed to work for him in an older version.

testbird:
@sketet Just in case, maybe the import/export function in the advanced settings can help you in the meantime.

[...]

PR #529, opened 2019-01-17
fix crash in imap

### LINKED ISSUE #527: sentboxThread kills Android app
- Delta Chat Core Version: 0.39.0 / Android-II 0.98.2 (Device with Android 9)
- Expected behavior:
- Actual behavior: DC for Android is crashing
- Steps to reproduce the problem:
- Screenshots:
- Logs:

Android app gets killed by:

`Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0xb in tid 29834 (sentboxThread)`

Full Log: [gist](https://gist.github.com/Ampli-fier/cfbf21866c7bf20cdeda0f1ec099981b)

r10s:
the problems was that `mailimap_fetch()` and `mailimap_uid_fetch()` may return a fetch_result != NULL on errors; this fetch_result. although set, is not valid then and must not be freed.

this pr checks if either an error is returned _or_  the fetch_result is NULL - in both cases, we set fetch_result to NULL so that it won't be freed again and regard the action as an error.

closes #527

### DIFF
--- a/src/dc_imap.c
+++ b/src/dc_imap.c
@@ -486,11 +486,12 @@ static int fetch_from_single_folder(dc_imap_t* imap, const char* folder)
 			read_cnt++;
 			if (!imap->precheck_imf(imap, rfc724_mid, folder, cur_uid)) {
 				if (fetch_single_msg(imap, folder, cur_uid)==0/* 0=try again later*/) {
+					dc_log_info(imap->context, 0, "Read error for message %s from \"%s\", trying over later.", rfc724_mid, folder);
 					read_errors++; // with read_errors, lastseenuid is not written
 				}
 			}
 			else {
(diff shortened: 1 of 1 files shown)


r10s:
@Ampli-fier thanks, this helps hunting the bug :)

Ampli-fier:
Using the preliminary (app) version under same conditions where the crash occured and the behaviour ist stable.

r10s:
@Ampli-fier thanks for testing - if someone else want to test out this exact bug, here is the app with the fix https://share.riseup.net/#GP2Pna3ax8oAbNYCJyKgaw

i'd like to check out some more things wrt to #156 before merging this pr to master, however.

Issue #413, opened 2018-11-01
Group chats sometimes split themselves and appear as a new group chat


Jikstra:
I don't know why but sometimes this happens, I think it's not even the same for desktop & android, but not sure about that.

r10s:
yes, this is the same between android and desktop and is a core problem.

the reason are the "ad-hoc" groups that are not created by deltachat, do not have a group-id and are identified mainly by the list of members.  

if the memberlist changes, this results in a new group, which is bad. however, i do not have a simple solution for this. allowing _adding_ members without creating a new ad-hoc group may mitigate the problem but probably not solve it. however, it may be a first step. not sure.

r10s:
okay, we just noticed this also for a "pure" delta chat group (in kyiv, "delta xi background org"), however, the reason for this still needs to be clarified.

hpk42:
in a verified group (bb33)  just got a split view which happened as follows:

- added a new verified member to the group from Fdroid-deltachat (core 0.20.0 IIRC), group has six members afterwards
- started desktop 1.7.4  (using core 0.24.0) and send a message with content "X" to the group 
- on F-droid-deltachat i get my own message in an split, non-verified opportunistic ad-hoc group of six people (so i don't think that desktop missed the member added -- but hard to know for sure because there is no "show original message" feature). the group name is "Chat: bb33 org: X" 
- on @angelo-fuchs 's dev and fdroid versions the split view does *not* appear -- all messages are properly listed in the one verified group, no splits.

r10s:
@hpk42 thanks for figuring out a (hopefully :) reproducible way. 

i will try to reproduce this later this way.

hpk42:
Not sure it's related, but when i change the group name with dev (!decentral), on desktop i now have two chats ... also there is some subject-mangling going on with android/core interactions as the subject names suddely appear in the UI with their mime-encoding.

Jikstra:
Same for me, i also saw many times that this happend on group name changes.

r10s:
not sure why the group name is relevant, however, part of the problem is that for any reasons the non-ascii-characters are not decoded on desktop; i also noticed this for "normal" names as "björn" :)

seems as if iconv is needed, cmp. #468 - unfortunately, the core compiles nevertheless but does not decode special chars.

r10s:
the commit above should fix the unwanted splits, however, there are still the following issues with groups:

[...]

PR #536, opened 2019-01-25
treat SELF-peer as being verified to avoid group-splits

r10s:
this pr modifies [check_verified_properties()](https://github.com/deltachat/deltachat-core/compare/selfpeer?expand=1#diff-1d2adc03d043975786fd5caef8655673R492) and skips the peerstate check if the message was sent by SELF.

this fixes potential group-splits in a multi-device setup as the SELF-peerstate is not really tracked nor needed.

as an alternative, we could also track the peerstate for SELF and leave the test as id, however, this would affect much more code and is not really needed afaik.

tackles #413


hpk42:
looks reasonable.

PR #542, opened 2019-01-26
[wip] soft-delete contacts when there are references

r10s:
done: scale down origin on contact deletion with references

missing: upscaling origin on incoming messages, might be a bit tricky as the chat_id currently implies the contacts are fine. hm.  
but maybe this is fine - contacts are not recreated on incoming messages then but _only_ when manually edited again (mainly when a name is given),

tackles https://github.com/deltachat/deltachat-android/issues/663


r10s:
stale pr, should be updated in core-rust.

PR #553, opened 2019-01-29
Add emailaddress to http autoconfig request

Ampli-fier:
This change is targeting #545.

Partially reverting [this commit](https://github.com/deltachat/deltachat-core/commit/a128a0ed694473867998e0d0776446f53b3e2ce4?diff=unified).

### DIFF
--- a/src/dc_configure.c
+++ b/src/dc_configure.c
@@ -792,7 +792,7 @@ void dc_job_do_DC_JOB_CONFIGURE_IMAP(dc_context_t* context, dc_job_t* job)
 		}
 
 		if (param_autoconfig==NULL) {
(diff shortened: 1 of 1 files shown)


r10s:
@Ampli-fier thanks for pushing this forward. 

one more comment: why not adding the address also to this line? https://github.com/deltachat/deltachat-core/pull/553/files#diff-f7047a1121c8ed04795e159497eb62cfR802

Ampli-fier:
a) It seems this works without supplying the mail address, see [here](https://developer.mozilla.org/en-US/docs/Mozilla/Thunderbird/Autoconfiguration#Configuration_server_at_ISP)
b) I cannot test/verify this

PR #554, opened 2019-01-30
Small etpan/osx tweaks

### LINKED ISSUE #512: Some meson pro should review the mac modifications
Some things i think are wrong
- [ ] I'm using `host_machine.system()` instead of `target_machine.system()` which will probably fail on cross compiling setups (not really important for the moment)
- [ ] We shouldn't do a libetpan on mac as it will fail anyway, maybe forbid it?

Probably some more stuff

flub:
- Use build_machine as we don't support any form of cross-compilation
  yet.

- Do not try to build the fallback etpan on osx.

Fixes #512


Jikstra:
Cool :+1:

r10s:
+1 - and welcome back, @flub :)

PR #559, opened 2019-02-02
avoid group-splits on verification problems

### LINKED ISSUE #413: Group chats sometimes split themselves and appear as a new group chat

r10s:
in verified groups, only encrypted messages encrypted and signed with out-of-band-verified keys are allowed. if a given message does not fulfill these needs, in the past, the message was shown in a parallel ad-hoc group. however, this results in some of ux problems, eg. missing overview, weired titles, and it was not clear what was going wrong.
if someone replies to the splitted group, things get even more worse :)

this pr tries to fix the problem by showing the reason of the problem in the original group; the original message text is only available dc_get_msg_info().
this way, replies go still to the same group and it is more clear who causes the problem and how it can be fixed.

closes #413

### DIFF
--- a/src/dc_receive_imf.c
+++ b/src/dc_receive_imf.c
@@ -505,7 +505,7 @@ static int check_verified_properties(dc_context_t* context, dc_mimeparser_t* mim
 		dc_log_warning(context, 0, *failure_reason);
 
 	if (!dc_contact_load_from_db(contact, context->sql, from_id)) {
-		VERIFY_FAIL("Cannot verifiy group; cannot load contact.")
+		VERIFY_FAIL("Internal Error; cannot load contact.")
 		goto cleanup;
 	}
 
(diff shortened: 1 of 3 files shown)


src/dc_receive_imf.c
```diff
@@ -499,8 +500,12 @@ static int check_verified_properties(dc_context_t* context, dc_mimeparser_t* mim
 	char*            q3 = NULL;
 	sqlite3_stmt*    stmt = NULL;
 
+	#define VERIFY_FAIL(a) \
+		*failure_reason = dc_mprintf("%s. See \"Info\" for details.", (a)); \
+		dc_log_warning(context, 0, *failure_reason);
+
 	if (!dc_contact_load_from_db(contact, context->sql, from_id)) {
-		dc_log_warning(context, 0, "Cannot verifiy group; cannot load contact.");
+		VERIFY_FAIL("Cannot verifiy group; cannot load contact.")
```
hpk42:
```suggestion
		VERIFY_FAIL("Internal Error; cannot load contact.")
```

src/dc_receive_imf.c
```diff
@@ -511,20 +516,20 @@ static int check_verified_properties(dc_context_t* context, dc_mimeparser_t* mim
 	if (from_id!=DC_CONTACT_ID_SELF)
 	{
 		if (!dc_apeerstate_load_by_addr(peerstate, context->sql, contact->addr)
-		 || dc_contact_is_verified_ex(contact, peerstate) < DC_BIDIRECT_VERIFIED) {
-			dc_log_warning(context, 0, "Cannot verifiy group; sender is not verified.");
+		 || dc_contact_is_verified_ex(contact, peerstate) != DC_BIDIRECT_VERIFIED) {
+			VERIFY_FAIL("The sender of this message is not not verified.")
```
hpk42:
```suggestion
			VERIFY_FAIL("The sender of this message is not verified.")
```

src/dc_receive_imf.c
```diff
@@ -511,20 +516,20 @@ static int check_verified_properties(dc_context_t* context, dc_mimeparser_t* mim
...
 			goto cleanup;
 		}
 
 		if (!dc_apeerstate_has_verified_key(peerstate, mimeparser->e2ee_helper->signatures)) {
-			dc_log_warning(context, 0, "Cannot verifiy group; message is not signed properly.");
+			VERIFY_FAIL("The message is not authorized.")
```
hpk42:
```suggestion
			VERIFY_FAIL("The message was sent with non-verified encryption.")
```

hpk42:
apart from the two minor suggestions looks fine to merge!

Issue #567, opened 2019-02-08
Sometimes messages are displayed twice


nicodh:
- Platform: mac
- Delta Chat Desktop Version: 0.100.0-PREVIEW
- Expected behavior: each message should only be displayed once
- Actual behavior: some messages are displayed twice
- Steps to reproduce the problem: Not sure yet, might be related to starting, stopping several times from command line?

I can see in the sqlLite database that there are multiple entries for messages with the same server_uid in the msgs table

<img width="823" alt="duplicate-messages" src="https://user-images.githubusercontent.com/1428457/52473581-77697700-2b96-11e9-8681-5a5be955b491.png">

<img width="748" alt="sql-lite-view" src="https://user-images.githubusercontent.com/1428457/52473700-c2838a00-2b96-11e9-8f1a-40c080a8f0f2.png">

r10s:
> This smells like a core problem.

yip, if the double entries are in the database, this seems to be a core problem, indeed.

would be great to get some hints about how to reproduce this, however :)

@nicodh thanks for filing this issue :)

nicodh:
So far I have no idea how this happened. I sometimes start DeltachatDesktop as "normal application" from the applications folder (downloaded as dmg) and sometimes via "npm start" in the deltachat root folder. But this did not cause any problems so far. And as you see in the screen, the first message from holger is duplicated, the second not...

So what would happen if both instances are open at the same time, writing to the same database. Could that result in a kind of "race condition"?
The we could just ignore this since there is no need that deltachat supports such a "developer" scenario...

r10s:
@nicodh thanks for explanations, yip sth. as a race condition is reasonable.

i assume this is a multi-device and/or multi-thread issue. 

seems as if, for whatever reason, messages pop up in different threads watching different folders _before_ they are inserted to the database (when the insert is done, this could not happen as the message-id is checked). reasons for that may be related to a second device moving the message.

if these assumptions go into the right direction, the doubled messages will differ in the server_folder and server_uid column - can you confirm this?

r10s:
i tried to explain in https://github.com/deltachat/deltachat-core/pull/569 what may have happened.

PR #569, opened 2019-02-09
[wip] avoid double messages 

r10s:
this pr tries to avoid double messages added to the database as observed in #567.

a message may be added to the database twice e.g. when two devices are running at the same time. 

the following example tries to explain this with the devices `mobile` and `desktop`:

1. `mobile.INBOX-thread` gets `msg1` from `INBOX`
2. `mobile.INBOX-thread` starts processing `msg1` as the Message-ID is not known
3. `desktop.INBOX-thread` also gets `msg1` from `INBOX`
4. `desktop.INBOX-thread` also starts processing `msg1` as the Message-ID is not known
5. `desktop.INBOX-thread` finishes processing the message before `mobile.INBOX-threads` continues, writes it to deskop-db and moves it on the server to the `DeltaChat` folder that is also watched by `mobile.DeltaChat-thread`
6. `mobile.DeltaChat-thread` gets `msg1` from `INBOX`
7. `mobile.DeltaChat-thread` starts processing `msg1` as the Message-ID is not known (`mobile.INBOX-thread` is still at 2.)
8. `mobile.DeltaChat-thread` finishes processing and writes `msg1` to mobile.db
9. `desktop.DeltaChat-thread` finishes processing and writes `msg1` to mobile.db

this is what one thing that may happen.

obviously, once a Message-Id is not known and processing is started, it should be blocked until really written to the database. a transaction may do the job, however, this may block too often unnecessarily, esp. when there are some checks between the Message-Id-Check and the INSERT.

### DIFF
--- a/src/dc_receive_imf.c
+++ b/src/dc_receive_imf.c
@@ -907,7 +907,6 @@ void dc_receive_imf(dc_context_t* context, const char* imf_raw_not_terminated, s
 	time_t           sent_timestamp = DC_INVALID_TIMESTAMP;
 	time_t           rcvd_timestamp = DC_INVALID_TIMESTAMP;
 	dc_mimeparser_t* mime_parser = dc_mimeparser_new(context->blobdir, context);
-	int              transaction_pending = 0;
 	const struct mailimf_field* field;
 	char*            mime_in_reply_to = NULL;
 	char*            mime_references = NULL;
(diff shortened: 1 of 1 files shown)


r10s:
closing this, the issue #567 is still open and points to the efforts done here.

as many things have changed in rust-core, not sure if this is still an issue. we'll see.

PR #574, opened 2019-02-11
WIP: changed from *.increation files to a separate dc_prepare_msg() call

VP-:
Added a new function `dc_prepare_msg()`. It is called with the same arguments as `dc_send_msg()` before. Then, `dc_send_msg()` is called later with the parameter `chat_id` set to `0` to actually create an SMTP job.

`dc_send_msg()` can still be called with a valid `chat_id` directly, like before.

### DIFF
--- a/src/dc_chat.c
+++ b/src/dc_chat.c
@@ -2176,7 +2176,7 @@ cleanup:
 }
 
 
-static uint32_t send_msg_raw(dc_context_t* context, dc_chat_t* chat, const dc_msg_t* msg, time_t timestamp)
+static uint32_t prepare_msg_raw(dc_context_t* context, dc_chat_t* chat, const dc_msg_t* msg, time_t timestamp, int state)
 {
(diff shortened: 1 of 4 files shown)


src/dc_chat.c
```diff
@@ -2473,21 +2431,126 @@ uint32_t dc_send_msg(dc_context_t* context, uint32_t chat_id, dc_msg_t* msg)
...
+ * e.g. the recoding of a video is not yet finished. Or the user can even forward
+ * the message with the file being still in creation to other groups.
+ *
+ * Files being sent with the increation-method must be placed in the
+ * blob directory, see dc_get_blobdir().
+ * If the increation-method is not used - which is probably the normal case -
+ * dc_send_msg() copies the file to the blob directory if it is not yet there.
+ *
+ * Example:
+ * ~~~
```
r10s: lower fence is missing

PR #575, opened 2019-02-12
Tell tox to pass LD_LIBRARY_PATH for testing of local code during development

VP-:
Allows to test locally-built code without installing it in /usr/local first.


hpk42:
thanks!

PR #579, opened 2019-02-19
probe imap STARTTLS on ports 993 and 143

### LINKED ISSUE #552: Improve automatic configuration
Delta Chat autoconfiguration seems to be smart enough, sadly it is a bit broken.
My server doesn't provide an autoconfig xml but it uses well know addresses and ports, so when I input "myusername@nauta.cu" and my password and leave empty all other fields, Delta Chat automatically try to use my email as user and "imap.nauta.cu"/"smtp.nauta.cu" for the IMAP/SMTP server, also the port is automatically "detected" based on the protocol and security method used, but the only thing that is broken is the IMAP Security setting(and probably SMTP Security also), I have to manually set it to STARTTLS, since Delta Chat only tries SSL and shows a toast with an error of ssl connection failed "imap.nauta.cu:993" instead of trying STARTTLS. This is the only thing preventing users to have an awesome user experience and providing only email and password without having to open advanced settings and filling all field since they never gonna guess that only disabling Security will work.
This have nothing to do with autoconfig xml files which is not provided by this server, that is only needed if the server use not well know ports  imo, and a MUA like thunderbird is able to automatically detect this settings without the autoconfig xml, as shown in the screenshots.

- Platform (android/blackberry/anbox): Android 6.0
- Device: BLU
- Delta Chat Version: 0.100
- Screenshots:

auto-config

autoconfig-results

r10s:
this pr improves the autoconfiguration by probing STARTTLS on port 993 after probing SSL on 993.

if both fail, both are retried with using only the local part of the mail as the login name.

closes #552

### DIFF
--- a/src/dc_configure.c
+++ b/src/dc_configure.c
@@ -928,32 +928,49 @@ void dc_job_do_DC_JOB_CONFIGURE_IMAP(dc_context_t* context, dc_job_t* job)
 	PROGRESS(600)
 
 	/* try to connect to IMAP - if we did not got an autoconfig,
-	we do a second try with the localpart of the email-address as the loginname
-	(the part before the '@') */
-	{ char* r = dc_loginparam_get_readable(param); dc_log_info(context, 0, "Trying: %s", r); free(r); }
+	do some further tries with different settings and username variations */
+	for (int username_variation=0; username_variation<=1; username_variation++)
+	{
+		// probe given settings, SSL/993 by default
+		{ char* r = dc_loginparam_get_readable(param); dc_log_info(context, 0, "Trying: %s", r); free(r); }
+		if (dc_imap_connect(context->inbox, param)) {
+			break;
+		}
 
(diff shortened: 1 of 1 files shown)


adbenitez:
what about of plain in 143??

r10s:
> what about of plain in 143

well, now as the logic with the username is simplified, we could rather easily this test.

however, i am not sure if there are many real-life-providers out there that do support none of STARTTLS and SSL on 993.

might also be a question of security when accepting PLAIN as a default or as a fallback, not sure.

adbenitez:
Nauta server support STARTTLS and plain, both on 143

r10s:
afaik also posteo (well-known german provider with focus on privacy) uses STARTTLS on port 143, so it makes probably some sense to test this as well. however, i would not test for PLAIN.

adbenitez:
I think it is ok without PLAIN
 :+1:

PR #580, opened 2019-02-20
simplify moving of dc messages

r10s:
leaving delta chat messages in the inbox to keep threads together seems to scare users more that tearing threads.

this pr moves messages with a `Chat-Version:` header to the DeltaChat folder. moreover, replies to dc messages are moved and the current move_state is tracked so that we can go for more complicated heuristics optionally at a later time.

this should result in a cleaner inbox and we can iterate from there, maybe add an option later or so.

this also mitigates the bcc-self story as these messages should also be moved in all cases and no more self-sent messages stay in the inbox.

this pr partly reverts https://github.com/deltachat/deltachat-core/commit/51bcf16c8c7d961d17eed723369b8a2e009fb784#diff-1d2adc03d043975786fd5caef8655673 and surrounding commits.

### DIFF
--- a/src/dc_move.c
+++ b/src/dc_move.c
@@ -3,133 +3,6 @@
 #include "dc_job.h"
 
 
-static dc_move_state_t determine_next_move_state(dc_context_t* context, const dc_msg_t* msg)
-{
-	// Return the next move state for this message.
-	// Only call this function if the message is pending.
-	// This function works with the DB, does not perform any IMAP commands.
-	dc_move_state_t res = DC_MOVE_STATE_UNDEFINED;
-	dc_msg_t*       newmsg = NULL;
-	dc_hash_t       handled_ids;
-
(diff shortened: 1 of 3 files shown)


adbenitez:
hopefully this fixes a bug: when a message is received if the connection fails when it is trying to move it to the DeltaChat folder, the message is never moved, check: https://github.com/deltachat/deltachat-core/issues/548

r10s:
might be, at least looking at the move logic [here](https://github.com/deltachat/deltachat-core/blob/master/src/dc_job.c#L101) and [here](https://github.com/deltachat/deltachat-core/blob/master/src/dc_imap.c#L1082) i do not see anything suspicious (maybe someone else does :)

hpk42:
looks good, i guess it'd be good to get an apk this week with this change ;)

r10s:
> looks good, i guess it'd be good to get an apk this week with this change ;)

yip ;)

PR #582, opened 2019-02-20
Add function to mark a message to be force-encrypted or force-unencrypted

chklump:
Add function to mark a message to be force-encrypted or force-unencrypted regardless of the heuristic to decide about the encryption.

This is related to [227 in the support forum](https://support.delta.chat/t/force-once-un-encrypted-to-directly-reach-a-classic-mua-after-having-had-a-chat/227) and also loosely related to #343 (I don't quite know what to reference here, because there are also corresponding GitHub issues and forum threads, respectively)

One could also think about extending the send command of the cmdline tool by a second optional parameter which triggers this, if desirable. (?)

### DIFF
--- a/src/dc_msg.c
+++ b/src/dc_msg.c
@@ -1094,6 +1094,34 @@ void dc_msg_set_text(dc_msg_t* msg, const char* text)
 
 
 /**
+ * Set if the message is sent out encrypted or unencrypted.
+ * This overwrites the heuristic which automatically desides if a message is encrypted or
+ * unencrypted. If set to encrypted here and encryption is not available later, the message
+ * will not be sent out at all.
+ * This does not alter any information in the database; this may be done by dc_send_msg() later.
+ *
+ * @memberof dc_msg_t
+ * @param msg The message object.
+ * @param encrypted 1=force-encrypt the msg, 0=force-unencrypt it
+ * @return None.
+ */
+void dc_msg_set_guarantee_e2ee(dc_msg_t* msg, int encrypted)
+{
+	if (msg==NULL || msg->magic!=DC_MSG_MAGIC) {
+		return;
+	}
+	if (encrypted) {
+		dc_param_set_int(msg->param, DC_PARAM_GUARANTEE_E2EE, 1);
+		dc_param_set_int(msg->param, DC_PARAM_FORCE_PLAINTEXT, 0);
+	}
+	else {
+		dc_param_set_int(msg->param, DC_PARAM_GUARANTEE_E2EE, 0);
+		dc_param_set_int(msg->param, DC_PARAM_FORCE_PLAINTEXT, DC_FP_ADD_AUTOCRYPT_HEADER);
+	}
+}
+
+
+/**
  * Set the file associated with a message object.
  * This does not alter any information in the database
  * nor copy or move the file or checks if the file exist.
--- a/src/dc_msg.c
+++ b/src/dc_msg.c
@@ -1095,7 +1095,7 @@ void dc_msg_set_text(dc_msg_t* msg, const char* text)
 
 /**
  * Set if the message is sent out encrypted or unencrypted.
- * This overwrites the heuristic which automatically desides if a message is encrypted or
+ * This overwrites the heuristic which automatically decides if a message is encrypted or
  * unencrypted. If set to encrypted here and encryption is not available later, the message
  * will not be sent out at all.
  * This does not alter any information in the database; this may be done by dc_send_msg() later.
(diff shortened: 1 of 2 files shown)


r10s:
thanks for pushing this forward.

i think we have to check this carefully wrt side-effects, however.

src/dc_msg.c
```diff
@@ -1093,6 +1093,34 @@ void dc_msg_set_text(dc_msg_t* msg, const char* text)
 }
 
 
+/**
+ * Set if the message is sent out encrypted or unencrypted.
+ * This overwrites the heuristic which automatically desides if a message is encrypted or
```
hpk42:
```suggestion
 * This overwrites the heuristic which automatically decides if a message is encrypted or
```

hpk42:
While the PRs changes by themselves look clean and nice, i am also a bit unsure about UX around this. 
https://support.delta.chat/t/force-once-un-encrypted-to-directly-reach-a-classic-mua-after-having-had-a-chat/227 and #343 do relate to this, but also do not give a complete UX picture, just state some wishes.   Usually it's better to change the core API in conjunction with real UI usage, from Android or Desktop.

src/dc_msg.c
chklump: Your right, of cause. Pushed commit to fix this.

chklump:
> Usually it's better to change the core API in conjunction with real UI usage, from Android or Desktop.

I planned to make a suggestion for deltachat-android similar to the way Signal does it (long press on the send button opens a menu, where you can choose "send" (the same as now), "send force-encrypted", "send force-unencrypted"). Should I make a suggestion for the implementation on the android-repo?

adbenitez:
a possible UX use for this is per chat encryption on/off???, for example I like encryption but to save some money I would like not to use encryption with some groups or 1x1 chats where most messages are shit-posting or a public irc room where the use of encryption makes no sense since it is public anyway. I must keep setting on and off encryption, current UX for this is pretty bad.

hpk42:
On Mon, Feb 25, 2019 at 18:29 -0800, Asiel Díaz Benítez wrote:
> a possible UX use for this is per chat encryption on/off???, for example I like encryption but to save some money I would like not to use encryption with some groups or 1x1 chats where most messages are shit-posting or a public irc room where the use of encryption makes no sense since it is public anyway. I must keep setting on and off encryption, current UX for this is pretty bad.

If we do encryption optioning on a per-chat basis then this PR's API
would likely not be needed.  

I sympathize introducing encryption options on a per chat basis.
however, there are questions/issues:

would everybody in an existing group have to accept one's changes to a chat's encryption settings? 

Or do we only allow encryption option changes at group-creation time? 

If no-encryption is selected, do we still send Autocrypt headers, or if not ...
how could we prevent recipients from interpreting this such 
that we can't do encryption also in other chats? 

Maybe @chkhlump's idea to go the signal route with a long-click
on the send button and switch between "force-encryption" and "force-cleartext" 
is workable and could be extended by stickyness -- the default is what was
forced the last time. This would still send autocrypt headers but would not
encrypt.  It might need an extension to Autocrypt, in that recipients reply
cleartext if they got a cleartext message ... 

All of these considerations touch UX, impl and autocrypt spec questions
which is why it hasn't just been quickly introduced yet ;) 

And which is why i think we can close this PR for now 
and rather drive this from the Android or Desktop repo. 

For a good example of this UX/UI-driven API change approach, 
see https://github.com/deltachat/deltachat-android/pull/742

adbenitez:
dear hpk, setting encryption on/off on a group is an user decision, and have the same effect as if user set encryption off globally, currently if an user set encryption off all group will had encryption disabled regardless if other members like the idea or not, anyway having an option(at creation time) in a group to declare that that group should always be encrypted is really useful and avoids the issues we currently have when an member of the group disables encryption.

hpk42:
dear adb -- i think we are both veering off-topic a bit for this PR ;) 
the PR under discussion is about introducing core API for setting
force-encrypt/cleartext on a per-message basis.  I am closing this
PR and suggest we take it from the forum and/or android issues
to continue discussion. 

the problem with turning encryption off for a group is its effect
on the relatively independent Autocrypt logic that does not consider
threads, chats or context. This means that switching off encryption
in one group could lead to switching it off in lots of other groups. 
I am sure there are ways to go about it and like i said, i sympathize
with doing this on a per-chat basis. 

Please also note that making encryption almost invisible is
a core feature of Delta Chat. So when we introduce UX for
changing this we want to do it carefully, considering related topics ---
and note that there are several people outside of Cuba who like
to discourage using Delta Chat because it's not strict about
enforcing e2e encryption ... :)

hpk42:
hello @chklump -- let's reconsider this change when we know better on the UX side what to go for.

hpk42:
hello @chklump -- let's reconsider this change when we know better on the UX side what to go for.

adbenitez:
pretty good points, sorry for the offtopic coment

Issue #427, opened 2018-11-08
avoid late generation of mime-structure


r10s:
currently, the mime-messages are generated from the database with the plain-text information just before sending (let us call this **late-mime-generation**):

the text and other information to sent is inserted to the database, we get a database-id and this database-id is added to a job-table then. as soon as the job is executed, the mime-structure is generated)

while this works usually well, this has some _downsides_:

- if the message could not be sent, the mime-generation is required on every retry; in some tests, we measured 500 ms or so for this
- if the message should be encrypted, we need an extra, message-depending-flag to force encryption on every retry (due to autocrypt the state may change)
- we cannot sent out a simple mime-message without having a full record in the database (normally, this is not required, but it may get handy here and there)
- the main _advantage_, however, is that this approach requires nearly no extra space on sending.

the alternative is **early-mime-generation**:

- advantages/disadvantages are inversion from above

however, when it comes to _single-sided ephemeral-messages_, early-mime-generation may be a MUST and not an option. 

the idea of _single-sided ephemeral-messages_ is that they do not stay on the device longer than a given amount of time. if this time is 0, however, we get the problem that the network may not be available just in that moment - and we cannot try resending later without breaking the promise of not saving the message.

this dilemma could be solved by encrypting the message to sent only to the recipients, and not to the own key, which, however, required the job-table being more independent from the message-table.

however, there are still open questions:

- in a multi-device-setup, this message won't reach the other devices. this is not necessarily a problem, however, it must be indicated to the user in a way
- these messages cannot be used in the SELF chat then

hpk42:
great writeup ... few comments ... 

On Thu, Nov 08, 2018 at 14:58 -0800, björn petersen wrote:
> the alternative is **early-mime-generation**:
> 
> - advantages/disadvantages are inversion from above

i think in flaky networks early-mime generation allows
to do the fastest push out to the SMTP server.
In the late-mime generation time is spent constructing
a mime message (which in the case of longer files may be substantial?)

also i guess the core can with early-mime generation more easily 
support bots or other "compose-mail-views" preparing the mime 
structure differently than currently. 

> [need to delete message on sending device ...]
> this dilemma could be solved by encrypting the message to sent only to the recipients, and not to the own key, which, however, required the job-table being more independent from the message-table.
> 
> however, there are still open questions:
> 
> - in a multi-device-setup, this message won't reach the other devices. this is not necessarily a problem, however, it must be indicated to the user in a way

[...]

PR #585, opened 2019-02-22
Changed from *.increation files to a separate dc_prepare_msg() call

VP-:
This is a re-creation of PR #574 because GitHub doesn't seem to like deleting/re-creation of branches after a rebase. I'll try a force-push next time.

The last commit also reverts PR #575 because adding `LD_LIBRARY_PATH` to `passenv`in `tox.ini`doesn't seem to be necessary after all. No idea what was broken when I tried it for the first time. If you want this as a separate PR, just drop a comment.

This is the first part for issue #427. It adds a new function `dc_prepare_msg()` which is called instead of creating an `*.increation` file and  calling `dc_send_msg()`. Then, when the actual file is ready, `dc_send_msg()` is called with `chat_id` set to `0`. The default use case with a non-zero `chat_id` continues to work as before.

The PR introduces a new message state `DC_STATE_OUT_PREPARING`for messages between the calls to `dc_prepare_msg()` and `dc_send_msg()`, and a new message parameter `DC_PARAM_PREP_FORWARDS` for the list of forwarded copies which were created while the original was still not ready. When `dc_send_msg()` is called on the original, the parameter is used to call `dc_send_msg()` on all forwarded copies (and recursively on all copies of copies).

One possible problem might be that `DC_STATE_OUT_PREPARING` has the value 21, while `DC_STATE_OUT_PENDING` has the value 20. This will be the first time that message go from a higher-numbered state to a lower-numbered one. If there are any inequality comparisons  with states, they might break now. Value 19 is already used by `DC_STATE_OUT_DRAFT`, and using an even lower value would still mean that the numeric value decreases when messages go from `…_DRAFT` to `…_PREPARING`.


r10s:
> One possible problem might be that `DC_STATE_OUT_PREPARING` has the value 21, while `DC_STATE_OUT_PENDING` has the value 20.

maybe we can define DC_STATE_OUT_PREPARING as 18, so the order would be PREPARING -> DRAFT -> PENDING -> DELIVERED -> MDN_RCVD which makes some sense as also drafts may be prepared theoretically and from PENDING on the order is clearer.

wrt inequality comparison: i've found only one in dc_msg_is_sent().

apart from that, at a first glace the pr looks good to me, however, i cannot say anything to the python stuff.

have you tested this already a bit? i'd also like to test it before merging, however, i would do this probably after the next playstore release.

VP-:
> have you tested this already a bit?

That was the point of the Python stuff. The new file `test_increation.py` contains the test which I used to test the code.

src/dc_chat.c
```diff
@@ -2517,21 +2475,137 @@ uint32_t dc_send_msg(dc_context_t* context, uint32_t chat_id, dc_msg_t* msg)
...
+ * ~~~
+ * dc_msg_t* msg = dc_msg_new(context, DC_MSG_IMAGE);
+ * dc_msg_set_file(msg, "/file/to/send.jpg", NULL);
+ * dc_send_msg(context, chat_id, msg);
+ * ~~~
+ *
+ * @memberof dc_context_t
+ * @param context The context object as returned from dc_context_new().
+ * @param chat_id Chat ID to send the message to.
+ *     If dc_prepare_msg() was called before, this parameter must be 0.
```
r10s: why not allow passing the chat_id even in this case?
imo this would provide a clearer api with less confusion.

i think we can check the prepare-state using `msg->state==DC_STATE_OUT_PREPARING` (as we already do) and if the state is differently, we assume, the message should just be sent as in the past.

and if the given chat_id differs from the chat_id the message was preprared for, maybe just call this an error.

VP-: According to the docs of `dc_msg_t`:

> The message object is not updated.

Although it is not strictly true, since the `id`, `chat_id` and `param` are updated by `dc_send_msg()`, `state` is indeed not changed (neither before nor after this PR). The check for `msg->state` in the code is actually a bug and passes the tests only because the Python bindings always return a new object loaded from the DB after the C call returns.

Theoretically, we should either change the "not updated" part of the doc and clarify *what* and *when* we change, or stop modifying `dc_msg_t` parameters. Depending on that decision (not necessarily waiting until the decision is fully implemented) we can either use `msg->state` instead of `chat_id`, or remove the check for `state`. And/or use something completely different to tell whether `dc_prepare_msg()` was already called (worst case: have `dc_send_msg()` look up the state in the DB).

r10s: > The message object is not updated.

this is more a warning that if you have a dc_msg_t object and the state changes in the database, eg. by receiving a read-receipt, the message objects are not updated.
maybe we should reword this part, however.

but i think it is fine to set the id and the state of a dc_msg_t object on sending - as mentioned, it already happens for some other field and the object is not declared as `const`.

@VP- apart from that: is this pr ready to merge from your point of view? i'd like to get it to master this week.

r10s:
@VP- \o/

PR #586, opened 2019-02-23
tone down gossip

r10s:
the aim of this pr is to add the [Autocrypt-Gossip](https://autocrypt.org/level1.html#key-gossip) headers less often without limiting its functionality too much.

for this purpose, the following changes are done:

- each chat has a timestamp marking the time of the last outgoing gossiped keys (`gossiped_timestamp`)
- `gossiped_timestamp` defaults to `0`
- `gossiped_timestamp` is reset (set to  `0`) for a single chat whenever the memberlist of the chat changes (never for 1:1 chats)
- `gossiped_timestamp` is reset for all chats on any incoming key changes (if needed, this can be optimized to reset `gossiped_timestamp` only for chats that really use the changed keys)
- finally, outgoing keys are gossiped only when `gossiped_timestamp` for a chat is `0` or older than 48 hours.

this saves about 2.5k per recipient per message. so already in small groups with 3 members a single _encrypted_ text messages saves about 5k and is now about 6k in size. in _larger_ group, this obviously saves some 10k per message.

for unencrypted messages, this change has no effect as keys are not gossiped at all there.

tl;dr - keys are gossiped on changes on keys or memberlist and once every 48 hours.

### DIFF
--- a/src/dc_apeerstate.c
+++ b/src/dc_apeerstate.c
@@ -197,6 +197,10 @@ int dc_apeerstate_save_to_db(const dc_apeerstate_t* peerstate, dc_sqlite3_t* sql
 		stmt = NULL;
 	}
 
+	if ((peerstate->to_save&DC_FORCE_REGOSSIP) || create) {
+		dc_reset_gossiped_timestamp(peerstate->context, 0);
+	}
+
 	success = 1;
 
 cleanup:
@@ -316,7 +320,7 @@ int dc_apeerstate_init_from_header(dc_apeerstate_t* peerstate, const dc_aheader_
 	peerstate->addr                = dc_strdup(header->addr);
 	peerstate->last_seen           = message_time;
 	peerstate->last_seen_autocrypt = message_time;
-	peerstate->to_save             = DC_SAVE_ALL;
+	peerstate->to_save            |= DC_SAVE_ALL | DC_FORCE_REGOSSIP;
 	peerstate->prefer_encrypt      = header->prefer_encrypt;
 
 	peerstate->public_key = dc_key_new();
@@ -336,7 +340,7 @@ int dc_apeerstate_init_from_gossip(dc_apeerstate_t* peerstate, const dc_aheader_
 	dc_apeerstate_empty(peerstate);
 	peerstate->addr                = dc_strdup(gossip_header->addr);
 	peerstate->gossip_timestamp    = message_time;
-	peerstate->to_save             = DC_SAVE_ALL;
+	peerstate->to_save            |= DC_SAVE_ALL | DC_FORCE_REGOSSIP;
 
 	peerstate->gossip_key = dc_key_new();
 	dc_key_set_from_key(peerstate->gossip_key, gossip_header->public_key);
@@ -358,7 +362,7 @@ int dc_apeerstate_degrade_encryption(dc_apeerstate_t* peerstate, time_t message_
 
 	peerstate->prefer_encrypt = DC_PE_RESET;
 	peerstate->last_seen      = message_time; /*last_seen_autocrypt is not updated as there was not Autocrypt:-header seen*/
-	peerstate->to_save        = DC_SAVE_ALL;
+	peerstate->to_save       |= DC_SAVE_ALL;
 
 	return 1;
 }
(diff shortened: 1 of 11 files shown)


src/dc_mimefactory.c
```diff
@@ -526,6 +527,13 @@ int dc_mimefactory_render(dc_mimefactory_t* factory)
 			}
 		}
 
+		// beside key- and member-changes, force re-gossip every 48 hours
+		#define AUTO_REGOSSIP (2*24*60*60)
+		if (chat->gossiped_timestamp==0
+		 || chat->gossiped_timestamp+AUTO_REGOSSIP < time(NULL) ) {
```
hpk42: maybe ```(time(NULL) - chat->gossiped_timestamp) < AUTO_REGOSSIP``` would be a more understandable way to describe the logic ;)

src/dc_apeerstate.c
```diff
@@ -316,7 +320,7 @@ int dc_apeerstate_init_from_header(dc_apeerstate_t* peerstate, const dc_aheader_
 	peerstate->addr                = dc_strdup(header->addr);
 	peerstate->last_seen           = message_time;
 	peerstate->last_seen_autocrypt = message_time;
-	peerstate->to_save             = DC_SAVE_ALL;
+	peerstate->to_save            |= DC_SAVE_ALL | DC_FORCE_REGOSSIP;
```
hpk42: isn't DC_SAVE_ALL almost implying DC_FORCE_REGOSSIP?  What would happen if we implicitely assume DC_FORCE_REGOSSIP with DC_SAVE_ALL?

hpk42:
a very nice and clear PR ... so i am fine to merge it, but curious about your replies to my two in-code comments.

src/dc_apeerstate.c
r10s: when the prefer-encrypt or a peer changes, if an encryption is degraded because of missing autocrypt headers, or if a key is verified, DC_SAVE_ALL is needed but DC_FORCE_REGOSSIP is not needed.

maybe it makes sense to force an REGOSSIP also in these cases. not sure, the idea to keep it separated was to save some traffic.

src/dc_mimefactory.c
r10s: do you mean `(time(NULL) - chat->gossiped_timestamp) > AUTO_REGOSSIP` ?

these things are hard ... ;)

hpk42:
On Mon, Feb 25, 2019 at 20:30 +0000, björn petersen wrote:
> r10s commented on this pull request.
> 
> > @@ -316,7 +320,7 @@ int dc_apeerstate_init_from_header(dc_apeerstate_t* peerstate, const dc_aheader_
>  	peerstate->addr                = dc_strdup(header->addr);
>  	peerstate->last_seen           = message_time;
>  	peerstate->last_seen_autocrypt = message_time;
> -	peerstate->to_save             = DC_SAVE_ALL;
> +	peerstate->to_save            |= DC_SAVE_ALL | DC_FORCE_REGOSSIP;
> 
> when the prefer-encrypt or a peer changes, if an encryption is degraded because of missing autocrypt headers, or if a key is verified, DC_SAVE_ALL is needed but DC_FORCE_REGOSSIP is not needed.

doesn't sound like an often occuring event if it's about pure DC-to-DC conversations. 
and after 2 days we anyway re-gossip ... 

> maybe it makes sense to force an REGOSSIP also in these cases. not sure, the idea to keep it separated was to save some traffic.

so i indeed think it's better to simplify and remove DC_FORCE_REGOSSIP, 
less room for bugs, less things to think about ;) 

btw, is there a "DC_SAVE" or why is there an "ALL"? ;)

r10s:
there is a DC_SAVE_TIMESTAMPS to save the autocrypt-timestamps only without touching the keys. the autocrypt-timestamp is changed on every single mail and i did not want the overhead. also i had a bad feeling touching the keys when not really necessary :)

so, every changed field that is not an autocrypt-timestamp , needs DC_SAVE_ALL - not only the keys but e.g. also prefer-encrypt values.

to the point: yes, i also think we can do a DC_FORCE_REGOSSIP implicitly when DC_SAVE_ALL is set. 

doing a commit.

r10s:
merging, we need to update the database also for the email-notification pr and i do not want to get these things messed up.

if there are open questions, we can handle them in a subsequent pr.

PR #589, opened 2019-02-25
[WIP] basic key generation using rpgp

dignifiedquire:
This expects rpgp to be installed using the following work flow

```
> git clone https://github.com/dignifiedquire/rpgp
> cd rpgp
> cd pgp-ffi
> make
```

After that run `meson && ninja` and things should build. 

Current  state: 
- ~~Tests are not happy with the public key for some reason~~
- ~~Fails to `free(user_id)`~~
- Stress tests are passing, and the cmdline cli boots up
- All required things are in `master` of rpgp

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,5 @@ python/src/deltachat/capi*so
 python/wheelhouse/
 
 .*.swp
+
+build
--- a/meson.build
+++ b/meson.build
@@ -44,6 +44,9 @@ math = cc.find_library('m')
 # this as well.
 zlib = dependency('zlib', fallback: ['zlib', 'zlib_dep'])
 
+# add rpgp
+rpgp = cc.find_library('rpgp')
+
 if not get_option('monolith')
   # Normal build, detect dependencies from pkg-config
   openssl = dependency('openssl', fallback: ['openssl', 'dep'])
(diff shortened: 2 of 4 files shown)


hpk42:
the makefile needs to use '-f' for removing: rm -f libpgp.h
so that it doesn't error out if the files does not exist.

hpk42:
followed the steps and at "make" time eventually it failed at compiling rand: https://gist.github.com/hpk42/78acd6415d580329a8821d96cd61e9ba

dignifiedquire:
hmm, that is odd, can you try running `cargo update && cargo clean && cargo test`? also can you make sure you have the latest commit on `extern-c` branch checked out?

dignifiedquire:
(I forced to that branch earlier today, so you might have an old version of it)

r10s:
cool :) great! happy to test this out.

maybe we can keep the old code a bit and use a `#ifdef RPGP` or so - then we can merge it and iterate from there without the need to have an android/whatever build-system ready just _now_.

hpk42:
the cargo-update,clean,test seems to progressed further but compile now fails elsewhere https://gist.github.com/hpk42/8e603e075e63493295c96cb15b0a5c59

dignifiedquire:
what rust version are you using?

hpk42:
rust-c --version gives "rustc 1.31.1 (b6c32da9b 2018-12-18)"

dignifiedquire:
closing in favor of #591

PR #591, opened 2019-02-28
Add experimental support rPGP 

dignifiedquire:
This expects rpgp to be installed using the following work flow

```
> git clone https://github.com/dignifiedquire/rpgp
> cd rpgp/pgp-ffi
> make install
```

After that run `meson && ninja` and things should build. 

Current  state: 
- Stress tests are passing, and the cmdline cli boots up
- All required things are in `master` of rpgp

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,5 @@ python/src/deltachat/capi*so
 python/wheelhouse/
 
 .*.swp
+
+build
--- a/README.md
+++ b/README.md
@@ -122,6 +122,11 @@ standard.  It is possible your system uses
 `/usr/local/lib/x86_64-linux-gnu` which should be auto-detected and
 just work as well.
 
+The library is build using netpgp by default (shipped in /libs).
+By adding `-Drpgp=true` to the build command, rpgp is used instead
+(https://github.com/dignifiedquire/rpgp,
+needs to be installed on the system or available as a lib)
+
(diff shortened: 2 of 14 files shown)


r10s:
changed the switch to DC_USE_RPGP - if defined, rpgp is used from the code side.

DC_USE_RPGP is not defined by default - systems that support it, should define it.

@flub can we get a switch for so easily to meson? the switch should define DC_USE_RPGP and do a `rpgp = cc.find_library('rpgp')` as in the current meson file. if undefined, everything should work as just now.
if that is too complicated, i would suggest to comment the changes in meson out for now; if needed they can be enabled then.

apart from the meson-thing, this pr looks good; it does basically nothing when DC_USE_RPGP is not defined.

the next step would be to get it run on android then :)

flub:
whoa, i pushed to this same branch.  magic

flub:
I realise there's no instructions on how to use that.  Invoke meson with `meson builddir -Drpgp=true` (or do the `meson configure -Drpgp=true` thing in an existing builddir)

dignifiedquire:
Update: all builds of `pgp-ffi` should be done using `--release --features nightly` and a nightly compiler ideally to ensure best perf + security available

r10s:
@dignifiedquire from my side, this can be merged :)

Issue #112, opened 2018-02-25
Read receipt  cleanup


Ampli-fier:
**Version info**: 

Delta Chat Core versions up to 0.15.0

**Observation**:

When deleting a message with Core on the Server, the corresponding _read/return receipt_ is not removed from the mailbox folder.

**Expected behaviour**:
Receipts are removed as well.

testbird:
I applied the user-impact labels, and chose "impairs: overall app usage" because I assumed users won't be very happy with an app that leaves a "mess" behind.

What do you think? Does the resulting rank reflect a true elevated relevance for users?

mogorman:
it was first thing my friend complained about when i got him to try it

Ampli-fier:
I'm happy with `severity 2`.

r10s:
> it was first thing my friend complained about when i got him to try it

but was the complaining about when you _deleted_ messages or in general about the large number of read receipts laying around in the folders normal MUAs. I think the latter is the larger problem.

mogorman:
I think his complaint is in regards to the fact that deltachat is inconsistent about keeping everything in its own folder. He also mentioned that some messages were being treated as chats, like short 1 liners.

r10s:
okay, thank you for the hint. the goal is to put all chat messages to the chats folder, in fact, there are issues about this and this has to be improved. do you know which version was used when testing this?

mogorman:
he was on 0.15.0 i believe, and i think only the receipts are currently remaining

r10s:
the read receipts are remaining in the INBOX and the other messages are moved to the Chats folder? this is interesting ...

mogorman:
the
Chat: Return receipt
messages.

webratte:
I don't know is my idea possible.
But still I will have it mentioned.

If a read receipt is arriving there will set the second chekmark.

Would it be possible to set a flag in the header to show other DC installations "this message is already readed" and then delete this read receipt.

Currently there are masses of read receipts in the deltachat folder. (One for every one liner 😉).

If you could the read receipt delete the way I think this would be great.

adbenitez:
please notice that not all email servers allow user defined flags, so the app would have to support the current behavior anyway to fall back if flags are not allowed

webratte:
I don't mean server flags.

I think something like a marker in the hidden header of the mail.

Something what say the other clients "this message have already received a read receipt ".

My english is really bad. 
I hope you will understand what I mean.

[...]

Issue #120, opened 2018-03-05
[consolidated] (auto-)archive / delete message history (support continuous daily usage)


testbird:
To avoid overgrowing amounts of old messages using all device "disk"-space, there should be an option to limit the retained history.

(The underlying functions will also be usefull when checking and hard-limiting storage usage on server and device, to support continuous usage.)

----

ToDo things:

Chat message selection
* [ ] **implement pinned/sticky/starred messages** (edit according to comments)
Allows to exempt messages from getting deleted by the following (automatic) bulk archiving and deleting.

Archiving and deletion management
* [ ] **implement managing and archiving of chats**
To keep chatting simple, manageable and search-able, only distinguish between active chats and  archived chats (no weekly, monthly etc. separate sub-archives).

  * [ ] core function to "delete all (non-sticky) older messages up to this one"
  * [ ] core function to "archive all (non-sticky) older messages up to this one"
  * [ ] UI option to "delete/archive old (non-sticky) messages up to this one" in *active* chats
  * [ ] UI option to "delete old (non-sticky) messages up to this one" in *archived* chats
  * [ ] UI button in chats to jump to first message (like existing button to jump to last message)
  * [ ] UI feature to auto-archive older messages from chats (default those exceeding a 100 messages count?)
  * [ ] UI feature to auto-archive stale chats (default to "no message since 6 weeks"?). If this gets to be the default, disabling this option for a chat could be called pinning / sticking / staring the chat to the active chat list.
  * [ ] the chat archive should get configuration options to
    * delete archived chats from server (and thus from other clients in a multi-client setup)
    * (?) move chats into a separate database (possibly encrypted if placed on external, removable storage that may have to be connected before access is possible, but would still allow to open and search the archive, and respond, if needed)
<br>

Storage Restrictions
* [ ] **Enable database compression (vacuum)**
     Actually free storage space when messages get deleted.
* [ ] **auto checks & deletes based on storage space limits (device & server)**
When the local filesystem, or the server space, is full, no more chat or email messages can be received.
  * [ ]  On incoming messages, let DC check the available space on the server and on the device, and when the free space falls below say 6% or 110MB, notify the user that DC will soon (below  5% or 100MB) have to start automatically deleting the oldest chat messages on the server or on the device.
  * [ ]  UI to manually set the storage usage limits (percentage and absolute) for automatic deletion of old archived and current messages.

Non-chat emails ("contact requests")

[...]

PR #595, opened 2019-03-02
[wip] purge messages from server after N days

r10s:
todo:

- [X] introduce an option
- [ ] purge messages from server after N days every N days
- [ ] maybe also delete read-receipts
- [ ] double-check the code
- [ ] test

to make things as safe as possible, we'll only purge message from the DeltaChat folder or, if moving is disabled, from the INBOX by a positive list (the known messages in the database)

targets #164 and maybe #112

### DIFF
--- a/src/dc_context.c
+++ b/src/dc_context.c
@@ -38,6 +38,7 @@ static const char* config_keys[] = {
 	,"mvbox_move"
 	,"show_emails"
 	,"save_mime_headers"
+	,"purge_server_days"
 	,"configured_addr"
 	,"configured_mail_server"
 	,"configured_mail_user"
@@ -518,6 +519,8 @@ static char* get_sys_config_str(const char* key)
  *                    also show all mails of confirmed contacts,
  *                    DC_SHOW_EMAILS_ALL (2)=
  *                    also show mails of unconfirmed contacts in the deaddrop.
+ * - `purge_server_days` = N=purge messages older than N days from server,
+ *                    0=do not purge messages from server (default)
  * - `save_mime_headers` = 1=save mime headers
  *                    and make dc_get_mime_headers() work for subsequent calls,
  *                    0=do not save mime headers (default)


adbenitez:
I guess that there will be an option to purge messages older than 0 days, so that all messages instead of moving to DeltaChat folder are purged directly and self-bcc are not sent at all, this will open possibilities for people more interested in good multidevice support than in syncing.

webratte:
Hi @r10s
If you are want be sure not delete wrong messeges, you simply could move this mails into the trash folder instead delete it completely.

What do you think?

adbenitez:
we need messages to be permanently deleted that is the whole point of this feature, the messages that will be deleted are the messages on DeltaChat folder, so what you suggests is to move messages to the Thrash folder instead of to the DeltaChat folder which is not much useful since the messages still remains on the server and we need to free space.

STPKITT:
Is there even an IMAP command which really deletes messages circumventing the trash?

adbenitez:
@STPKITT well you actually flag them as /Deleted and then issue expunge command, or something like that.

testbird:
Please implement this to be the server-limits and -cleanup part of the "auto-delete" background task from https://support.delta.chat/t/reduce-space-on-internal-storage-disk-full/132/24 (keeping in mind its later extension local-limits and -cleanup)

The -limits related details from #120 contain a nice storage space condition:

* [ ] auto checks & deletes based on storage space limits (device & server)
When the local filesystem, or the server space, is full, no more chat or email messages can be received.
  * [ ]  On incoming messages, let DC check the available space on the server and on the device, and when the free space falls below say 6% or 110MB, notify the user that DC will soon (below  5% or 100MB) have to start automatically deleting the oldest chat messages on the server or on the device. Non-chat emails should not be deleted from the server.

webratte:
With "location streaming"  comes a lot more mails which flooded the mailbox.

So delete older messages become more important.

testbird:
The problem of many meta-data messages mingled within regular messages won't get solved by a history limit.

Looks like #112 remains a different issue, that might best be solved with #644, using a "first see, first delete" approach that stores the updated state for the other clients using message flags or state messages.

webratte:
@testbird
I don't agree.
A history limit would help to save server storage.

I could set it e.g. to 30 days. Time enough to sync my clients.

People with really less storage (e.g. @adbenitez) and no need for sync, could set it e.g. to 1 day or so.

So purge older messages would definitely help.

adbenitez:
webratte actually I was expecting a set to 0 days where the messages are deleted instead of moving them after they are downloaded and the self-bcc copy isn't even send :) really I don't want them on the server at all

On April 21, 2019 1:12:23 AM GMT-04:00, webratte <notifications@github.com> wrote:
>@testbird
>I don't agree.
>A history limit would help to save server storage.
>
>I could set it e.g. to 30 days. Time enough to sync my clients.
>
>People with really less storage (e.g. @adbenitez) and no need for sync,
>could set it e.g. to 1 day or so.
>
>So purge older messages would definitely help.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/deltachat/deltachat-core/pull/595#issuecomment-485225509

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

adbenitez:
and when I say "me" I am talking of the Cuban use case of curse, not my particular preference it is just that we need to use 50 MB for a month with care

On April 21, 2019 1:12:23 AM GMT-04:00, webratte <notifications@github.com> wrote:
>@testbird
>I don't agree.
>A history limit would help to save server storage.
>
>I could set it e.g. to 30 days. Time enough to sync my clients.
>
>People with really less storage (e.g. @adbenitez) and no need for sync,
>could set it e.g. to 1 day or so.
>
>So purge older messages would definitely help.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/deltachat/deltachat-core/pull/595#issuecomment-485225509

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

testbird:
> I don't agree.
>A history limit would help to save server storage.
That's not the problem, and not what I wrote.
The problem is that a global history limit would not reduce or filter out the receipts etc. without also deleting the desired chat history.
But an separate syncing feature may fllter out the receipts etc. and reduce the server storage used by the chat history a lot.

webratte:
I think this can refined in a later version.

As a first step it would be enough to delete mails older then "n" days.

Or maybe directly after download if no sync is needed.

testbird:
Yeah, #112 is a separate issue, that's what I meant (besides that actually quota based deletes will be needed to guarantee free space, especially with small servers).

#112 is not solved with this issue, contrary to the suggestion in the report (and #112 is actually better solved by deleting receipts with syncing feature #644 rather than only with the corresponding message).

testbird:
Wait a second, this issue is about *server space*, right?

Then why would one want to delete messages after N days, instead of limiting the number of messages?
Small servers may fill up quickly with a wave of messages. And low-traffic but high-interest chats can vanish too quickly, if a short temporal value is used trying to control a size value. (dimension mismatch)

Maybe have a look at this (updated) discussion post (again),
https://support.delta.chat/t/reduce-space-on-internal-storage-disk-full/132/24
this issue would corresponds to a first version of the "auto-delete-server".

adbenitez:
testbird, people want to delete based on time because in general the older a message is, the less important it is for an user, and the less they will remember it even exists

testbird:
> testbird, people want to delete based on time because in general the older a message is, the less important it is for an user, and the less they will remember it even exists

Remember this is a server issue.
We need to look at a more complete picture. I'll try to recap.


It's  like this:
Peopple simply just see the chats.

And people also usually only want, and only need to deal with simple options to adjust the "history lengh" kept in chats. But this is NOT the "Max. number of messages kept on server", for which there should usually be no need for the user to change it.

The useful knobs for the user to shorten the list of messages in chats is rather "Limit number of (archived) messages (per chat)" or "Limit age of (archived) messages (per chat)" The latter would also completely delete important but low traffic chats, however.

Note that there is no requirement to configure a history limit. (There are users with huge server and device strorage.) However, an auto-delete-server/client is needed to avoid disk full errors, and avoids having to ship with a default history limit.)  Limits should be easily configured though, if desired, as a personal preference.

If you review the [options](https://support.delta.chat/t/reduce-space-on-internal-storage-disk-full/132/24) (I improved the sorting and wording.), I think the server setting "Max. number of messages kept on server, and thus synced accross devices (per chat)" would be a really advanced "admin" setting, needed for the mentioned special use-cases. And it makes sense for it to be an integer value, in order to make it easy for the clients to count sync the n messages, and detect those expiring at the end of the list.

webratte:
I don't understand you.
There are much very important things to do.

Why should the devs start complicated as you mentioned?

This issue is important.
But a relatively easily start (so I think, I can't coding) could be:

-Check the deltachat folder every e.g. monday for messages older then 6 month as a user preference

-Now delete this older messages without to have to check which chat is involved.

Of course, this can be defined if the devs onetime will have nothing more important things to do 😉

testbird:
If using an amount of days for expiration on the server, a second client can not know the number from the messages on the server, and can't distinguish between a delete to sync (delete locally) and a delete to keep (archive locally). So on some syncing devices, nothing may be kept (enter the local archive) even if configured.

webratte:
Sorry, I don't know what you mean...

testbird:
> every e.g. monday

That will look strangely erratic to users, not help in a week with a lot of spam messages, not free up enogh space, ... and has problems if device is off one day.

Simplest: do the check on every client with every received message. 

> without to have to check which chat is involved

We're talking about the server here, and keeping a Trashbin is useless to free up space. Users will complain: I come back from vacation and deltachat deleted all my important messages just because friend/group X had sent me a lot (or large) messages.

testbird:
Avoid sure problems when a better way is known.
Don't start in a direction if it is known to create more problems in the next steps than in solves "for now".

testbird:
And again, in the broader picture this issue could be seen as a first "auto-delete-server".

webratte:
>      

> 

> We're talking about the server here, and keeping a Trashbin is useless to free up space. 

We talking about DC (the App) will check the server and delete the older messages.

> Users will complain: I come back from vacation and deltachat deleted all my important messages just because friend/group X had sent me a lot (or large) messages.

Only if theay was more then 6 month in hollydays (for my example)

testbird:
> Only if theay was more then 6 month in hollydays (for my example)

See, particular examples are not working in general. Other users with few server space would have to reduce this purge setting a lot. And everything must still work. Half baked "features" can be detrimental.

webratte:
> See, particular examples are not working in general. Other users with few server space would have to reduce this purge setting a lot. And everything must still work. Half baked "features" can be detrimental.

Of course, mails should not deleted before it's downloaded.

adbenitez:
I also think that instead of dreaming with a "perfect" method, we start with something simple that "works", we are in a situation with 100MB of server space, users run out of storage and they blame Delta Chat, because **other email-chatting apps don't have this "error"** so we need something that just delete the messages after they are downloaded to DC is data base, so even starting with a simple option like that (keep on server or remove them after download) will be enough for an start, of curse deleting based in time would be also something good to have, my point here testbird:
lets first have something that at least "works" and that can leverage the server storage situations for users and then we could work in more complex/fancy stuff.

testbird:
I don't think anybody said to delete without the first client in the race downloaded the message. But the problem comes with later syncing devices, especially coming back after some time. 

>  start with something simple that "works",

Yes, sure. I am all for a auto-delete-server, even if starting only having the "fallback" for the case without server quoata.

My point is just that a time based limit for server deletes does not work for multi-client sync. (A mistake to avoid.)

A total max. number of chat messages on the server is a technical setting that can also work for syncing later I think.
For limiting the history, a much more usefull option for the user should be a simple separate "max. number per chat" (auto-delete-device).

webratte:
I think we should stop at this point our discussion and let the experts (not you and not I but the devs) do what they think it's a possible way.

Isn't it?

testbird:
OK.
Without support for detectable free space (quota), a FIFO defaulting to something like 250 messages should be ok to put a default active bound on the chat folder, because
* deletes require that one device is online anyway (hopefully not fully lost after one device downloaded),
* the multi-device cooperation rule "If the oldest message gets delted, assume an expiration, and keep the local copy" is simple and may work without requiring separate sync and coordination data and storage overhead.

webratte:
Should this issue reopened in rust-core?
I believe this is still a very important feature.

r10s:
> Should this issue reopened in rust-core?

this is not an issue, this is the unfinished try to fix an issue.

the belonging issue is still open and already moved to core-rust: https://github.com/deltachat/deltachat-core-rust/issues/512

PR #602, opened 2019-03-06
Update README.md, suggest using `sudo -E`

csb0730:
`sudo -E ninja install`

The -E parameter is necessary to get this commad working as expected.
Reason is because meson is installed for current user (pip) with given instructions and if we use sudo command python3 is not able to find installed package. sudo sees another environment as current user.
The -E command passed current environment to sudo user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ meson
 # run `meson configure` to see the options, e.g.
 #    meson configure --default-library=static
 ninja
-sudo ninja install
+sudo -E ninja install
 sudo ldconfig
 ```
 


r10s:
i'm not sure if the general advice of passing the flag `-E` to the `sudo` command is needed.
maybe it depends on how things are set up, eg. i cannot remember that the flag was needed for me.

maybe just add hint beside the raw command and let the user decide if this is needed?

sth. as `sudo ninja install # if you run into problems, the sudo-flag -E might help`

csb0730:
Yes, this is sufficient and the best procedure. I think, I can only describe what I see here at my own pc but don't know about the case at other installations.

But by the way @r10s : I see further difficulties while building core. For example I have problems with dependencies of libEtPan. I'm able to build (with some warnings, and only in using -Dmonolith=true) but at the end resulting command line tool doesn't show a prompt.

Shall I open a new issue for that? For example with title "build problems caused by dependencies or environment" ?

r10s:
well, in general, i've rarely seen any app or library without the one or the other build problem :)

r10s:
in general, i think the commands shown should work on a fresh installed debian or so.

csb0730:
I think I found the problem after deleting full builddir content and doing all new from scratch:

This is output part for libetpan:

```
Program libetpan-config found: YES (/usr/bin/libetpan-config)
Message: Using libetpan fallback...

|
|Executing subproject libetpan 
|
|Project name: etpan
|Project version: 1.8
|Native C compiler: cc (gcc 6.3.0 "cc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516")
|Dependency threads found: YES (cached)
|Dependency zlib found: YES (cached)
|Dependency openssl from subproject libs/openssl found: YES (cached)
|Dependency sqlite3 from subproject libs/sqlite found: YES (cached)
|Dependency libsasl2 from subproject libs/cyrussasl found: YES (cached)
|Compiler for C supports arguments -Wno-format-truncation -Wformat-truncation: NO
|Compiler for C supports arguments -Wno-misleading-indentation -Wmisleading-indentation: YES
|Compiler for C supports arguments -Wno-sometimes-uninitialized -Wsometimes-uninitialized: NO
|Compiler for C supports arguments -Wno-unused-function -Wunused-function: YES
|Build targets in project: 4
|
|Subproject libetpan finished.

```
==> it worked now with fallback :-)


After launching command:

```
myhome:~/ ... /cmdline$ ./delta test1.db
First time init: creating tables in "test1.db".
Opened "test1.db".
```
it needs more than a minute after line `Opened test1.db` appears until prompt appears!
I didn't had the patience to wait so long ;-)

So far I think problems are solved and I'm able to go on :-)

csb0730:
Maybe we should do following optimizations to cmdline tool so that everyone has a good user experience (UX) with it at once:
- don't run stress functions by default. Only activate them with a parameter.
- add a hint after start of cmdline tool that some functions are only done if threads are started and ask to start them. Or print a clear statement what to do.

r10s:
duckduckgoing a bit, i think the `-E` option is too special for a general advice. it may be restricted and it may even cause problems.

other environments may have different solutions. the readme is not the place to explain all kind of possible problems, it should be enough to show one "happy path" that generally works and shows the steps needed.

for special thinkgs, the issue tracker can be used/searched.

csb0730:
Hi @r10s 
it needs many hours to find out why build process isn't working with the described commands. I find it really necessary to do a comment about -E option at build command. This is not a nice to have, this is a requirement!
Maybe this differs relating linux installation you have installed, but in any case:
- The described commands to install the packages (pip) do it only for current user
- so sudo isn't able to find them and -E option is a requirement (except you are working as root?!).

IMO it's not ok to close this pr without leaving a comment in build commands.

You described yourself a the best way to do a comment (10 days ago)!

csb0730:
I mean: We do speak here about a comment (!), not a rule.

r10s:
sure, and we can add a comment, however, this pr suggests to add -E in general to the sudo command and therefore it was closed.

as mentioned, maybe we can add it as a comment.

r10s:
don't worry, i add the command directly to the readme :)
though i do not understand it fully, it might help :)

r10s:
@csb0730 added, thank you for your help :)

csb0730:
@r10s thanks a lot :+1: 

It will help and saves hours for anyone who runs into that problem :-)

PR #607, opened 2019-03-13
[wip] prefetch less headers

### LINKED ISSUE #598: check if prefetch of headers can be optimized
currently, we're prefetching the headers using `FETCH ENVELOP` https://tools.ietf.org/html/rfc3501#section-2.3.5 - if this also includes the Autocrypt-Headers, maybe we can optimize that (an Autocrypt-Header with an RSA-key is about 2.5K, an ECC-key is 0.5K - as this sums up to some K when prefetching 10 messages, even the 0.5K this sum up and it's probably worth the effort)

r10s:
this pr changes the prefetching of mails to a positive list of needed headers instead just using the ENVELOPE (which, however, did not include the Autocrypt header, but eg. complete To:-lists)

- [x] prefetch only Message-ID and Chat-Version
- [ ] testing, esp. deletion which uses the same prefetch structure

closes #598

### DIFF
--- a/src/dc_imap.c
+++ b/src/dc_imap.c
@@ -202,7 +202,7 @@ static char* unquote_rfc724_mid(const char* in)
 }
 
 
-static const char* peek_rfc724_mid(struct mailimap_msg_att* msg_att)
+static char* get_rfc724_mid(struct mailimap_msg_att* msg_att)
 {
 	if (msg_att==NULL) {
 		return NULL;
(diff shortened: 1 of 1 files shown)


testbird:
Can't the `Chat-Version` header be considered optional, with no need to be prefetched, if the Message-IDs already start with an identifiable string like `Mr.` or maybe something little less random like `E-Chat.`?

adbenitez:
dc will be coi compatible so it will start with coi$ I guess

r10s:
stale pr, should be updated in core-rust.

PR #611, opened 2019-03-19
Emit DC_EVENT_MSGS_CHANGED from dc_prepare_msg()

VP-:
To allow UIs to rely entirely on events do display messages, `dc_prepare_msg()` should emit events just like `dc_send_msg()`. Calling `dc_send_msg()` directly will emit the event only once, since the PREPARING state is skipped entirely.


r10s:
lgtm :)

PR #622, opened 2019-03-23
Use pkg-config to find rpgp library

Jikstra:


meson.build
```diff
@@ -111,8 +111,8 @@ endif
 if get_option('rpgp')
   # The rpgp pkg-config file is currently incorrect, so just try
   # linking it instead for now.
-  #rpgp = dependency('rpgp')
-  rpgp = cc.find_library('rpgp')
+  rpgp = dependency('rpgp')
+  #rpgp = cc.find_library('rpgp')
```
hpk42: shouldn't this remove the comment above?

Jikstra: @flub any ideas?

flub: yes, also no need to keep the cc.find_library(...) around in commented form, just delete it

flub: but whatever, let's merge this since ppl already fall over this on irc.  we can remove comments afterwards

PR #625, opened 2019-03-24
introduce automated building of deltachat python bindings …

hpk42:
…(as wheels) and upload them m.devpi.net/dc/BRANCHNAME


.scripts/buildwheel.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+set -e -x
+
+docker pull deltachat/wheel 
+rm -rf python/wheelhouse/*
+cd $TRAVIS_BUILD_DIR
+docker run --rm -it -v $(pwd):/io deltachat/wheel /io/python/wheelbuilder/build-wheels.sh
```
flub: heh, so you stuck with /io :wink:

.travis.yml
```diff
@@ -1,21 +1,18 @@
 sudo: required
```
flub: I swear the docs say you need:
```
services:
  - docker
```
But this appears to work??  I'd still add it though, better not to rely on undocumented behaviour I guess.

python/README.rst
```diff
@@ -6,34 +6,57 @@ This package provides bindings to the deltachat-core_ C-library
...
+with ``pip``.  For example, to install the latest "master" built::
+
+    pip install -i https://m.devpi.net/dc/master deltachat
+
+
+Installing bindings from source
+===============================
+
+If you can't use "binary" method above then you will need
+to `install the delta-core C-library <https://github.com/deltachat/deltachat-core/blob/master/README.md>`_ before you invoke the ``pip install deltachat`` command above.
```
flub: `pip install --no-binary deltachat` to avoid using the wheel and compile against the system libdeltachat.so.  This is useful in case pypi has an outdated libdeltachat.so because we forgot/got hit by a bus.

python/README.rst
```diff
@@ -6,34 +6,57 @@ This package provides bindings to the deltachat-core_ C-library
...
+
+.. note::
+
+    If you can help with provide a way to automate the building of wheels
+    for Mac or Windows, please provide a PR here or open an issue.
```
flub: How about mention to discuss on delta@codespeak.net or irc?  It's a lower barrier that having to create a full-fledged PR...

python/setup.cfg
```diff
@@ -2,5 +2,3 @@
 formats = sdist.tgz
 no-vcs = 1
 
-[bdist_wheel]
-universal = 1
```
flub: Oops, good catch!

PR #633, opened 2019-03-27
[wip] rework travis triggered jobs to easily run off-travis, remove build redundancy

### LINKED ISSUE #628: Simplify travis.yml / run tests within docker container
I suggest to move the tox and ninja invocations to run inside the docker image, instead of through travis.  This would remove a lot of hard-to-maintain/evolve travis-config which really is a duplication of what we do in the dockerfile.  The deploy part also would need to run inside the docker, so we might need to arrange a bit differently with the ssh-secrets.

hpk42:
fixes #628


hpk42:
closed in favor of #639

PR #634, opened 2019-03-28
more powerful dc_get_locations()

r10s:
this pr modified dc_get_location() so that `0` can be given as a contact id. together with the timespan and the chat_id, this function some powerful filtering options that can be used for both, global and chat map.

### DIFF
--- a/cmdline/cmdline.c
+++ b/cmdline/cmdline.c
@@ -302,8 +302,7 @@ static void log_msglist(dc_context_t* context, dc_array_t* msglist)
 }
 
 
-static void log_contactlist(dc_context_t* context, dc_array_t* contacts,
-                            uint32_t add_locations_for_chat)
+static void log_contactlist(dc_context_t* context, dc_array_t* contacts)
 {
 	dc_contact_t*    contact = NULL;
 	dc_apeerstate_t* peerstate = dc_apeerstate_new(context);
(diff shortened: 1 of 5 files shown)


nicodh:
Since it is possible now to get the locations for all chats (for global map), shouldn't we have a dc_array_get_chat_id method, to see which location was sent to which chat?

nicodh:
tested get_locations for branch - works as expected!

r10s:
@nicodh thanks again for merging and for testing :)

> shouldn't we have a dc_array_get_chat_id method, to see which location was sent to which chat?

i think it is currently not needed as it would always return the chat-id given to `dc_get_locations()`.

for messages bound to locations, however, you're probably interested in the whole message object anyway - which also contains the chat-id then:
```
dc_msg_t* msg = dc_get_msg(context, dc_array_get_msg_id(arr, index));
dc_msg_get_chat_id(msg)
```

nicodh:
But for the global view we don't pass a chat ID as said in the examples comment. So do we only show contacts in global view? Or should we "group" them by chats. I thought we would go for the latter...

r10s:
> Or should we "group" them by chats.

that's a point. although i would not target this filtering ux/ui wise right now, i do a pr to have the api complete :)

r10s:
@nicodh https://github.com/deltachat/deltachat-core/pull/636 :)

PR #639, opened 2019-03-29
replace travis with circle-ci, cleanup of build steps

### LINKED ISSUE #628: Simplify travis.yml / run tests within docker container

hpk42:
Travis is getting slower and was already hard to debug. 
So this PR: 
- moves from travis to Circle-CI
- has a ci-scripts/README.md which explains what is contained there 
- puts all build and CI scripts into "ci_scripts" directory 
- allows to run things locally 
- trims down the base Docker image where tests/docs/wheels are run from 4.4GB to 1.7GB
- fixes #628 

Using circle-ci also means that we can probably integrate Mac-OS build steps more easily. 

I am sure that my initial circle-ci config (helped by kind hints from @dignifiedquire ) can be improved.


ci_scripts/README.md
```diff
@@ -0,0 +1,52 @@
+
+# Continous Integration Scripts for Delta Chat
```
ralphtheninja: Continuous

ci_scripts/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+Continous Integration is run through CircleCI
```
ralphtheninja: Continuous

ci_scripts/README.md
```diff
@@ -0,0 +1,52 @@
...
+the hub for sharing Docker images::
+
+    docker pull deltachat/coredeps
+    docker pull deltachat/doxygen 
+
+or you can build the docker images yourself locally
+to avoid the relatively large download:: 
+ 
+    cd ci_scripts  # where all CI things are 
+    docker -t deltachat/coredeps docker-coredeps
```
ralphtheninja: `docker build -t ..`?

ralphtheninja:
Nice work holga!

ci_scripts/config.yml
```diff
@@ -0,0 +1,90 @@
...
+      - attach_workspace:
+          at: workspace
+      - run: ls -laR workspace 
+      - run: ci_scripts/ci_upload.sh workspace/c-docs workspace/py-docs workspace/wheelhouse
+
+  build_monolith_tests:
+    machine: True
+    steps:
+      - checkout
+      - run: docker pull deltachat/test7
```
flub: Is this the image you meant to pull?

ci_scripts/config.yml
```diff
@@ -0,0 +1,90 @@
...
+          command: |
+            echo 'export MESONARGS="-Dmonolith=true"' >> $BASH_ENV
+            echo 'export TESTS=1' >> $BASH_ENV 
+      - run: ci_scripts/ci_run.sh
+
+  build_static:
+    machine: True
+    steps:
+      - checkout
+      - run: docker pull deltachat/test7
```
flub: Same?

ci_scripts/config.yml
```diff
@@ -0,0 +1,90 @@
...
+      - run:
+          command: |
+            echo 'export MESONARGS="--default-library=static"' >> $BASH_ENV
+      - run: ci_scripts/ci_run.sh
+
+  build_forcefallback: 
+    machine: True
+    steps:
+      - checkout
+      - run: docker pull deltachat/test7
```
flub: And again

ci_scripts/docker-coredeps/deps/build_libetpan.sh
```diff
@@ -0,0 +1,19 @@
+
```
flub: I think you should really add #! lines to all these scripts

ci_scripts/docker-coredeps/deps/run_all.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+#
+# Build the Delta Chat C/Rust library
+#
+set -e -x
+
+# perform clean build of core and install 
+export NINJA_BUILD_DIR=.docker-corebuild
```
flub: A hidden directory is a rather weird choice here

meson.build
```diff
@@ -111,8 +111,8 @@ endif
 if get_option('rpgp')
   # The rpgp pkg-config file is currently incorrect, so just try
   # linking it instead for now.
-  rpgp = dependency('rpgp')
-  #rpgp = cc.find_library('rpgp')
+  # rpgp = dependency('rpgp')
```
flub: If we need to do this could we do this in a separate PR please? I'm not sure why you need it and would like to see if there's an other solution to whatever made you do this

ci_scripts/ci_run.sh
```diff
@@ -0,0 +1,20 @@
+# perform CI jobs on PRs and after merges to master.
+# triggered from .circleci/config.yml
+
+set -e -x
+
+export BRANCH=${CIRCLE_BRANCH:-test7}
+    
+# run doxygen on c-source (needed by later doc-generation steps).
+# XXX modifies the host filesystem docs/xml and docs/html directories
```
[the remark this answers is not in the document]
hpk42: i tried it but it seems that then "ninja install" fails inside the docker because it can't write to /usr/local

ci_scripts/config.yml
hpk42: no, fixing :)

ci_scripts/docker-coredeps/deps/build_libetpan.sh
hpk42: k

ci_scripts/docker-coredeps/deps/run_all.sh
hpk42: i don't want to interfere with the host system -- all this scripts run in docker and write to the host system, for now.

flub: you're still writing to the host though, only the user doesn't see it.  if you don't want to interfere then maybe check if the directory exists and if it does bail out with an error on the CI it shouldn't exist so will work?

ci_scripts/ci_run.sh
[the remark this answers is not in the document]
flub: does `sudo ninja install` work in the docker container?

meson.build
hpk42: removed here.

ci_scripts/docker-coredeps/deps/run_all.sh
hpk42: i am not up for doing the circle-ci cycles right now and suggest to do this later. In my testing and usage it seems to be fine.  We could use more of circle-ci to isolate things, i.e. not mount from the host -- i think circle-ci does let you start in the checkout but this checkout is a layer, not a direct mount. I might be mistaken.  However, that would mean not being able to run things locally and for now i prefer to be able to have quick cycles locally.  This spurious artifact is only a minor annoyance IMO.

ci_scripts/ci_run.sh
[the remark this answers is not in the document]
hpk42: "sudo" is command not found i think.

python/tox.ini
```diff
@@ -28,14 +38,13 @@ commands =
     rst-lint --encoding 'utf-8' README.rst
 
 [testenv:doc]
-usedevelop = True
 deps =
-    sphinx
+    sphinx==1.8.5
```
flub: What's the reason to lock this version?

ci_scripts/docker-coredeps/deps/build_rust.sh
```diff
@@ -0,0 +1,11 @@
+
```
flub: Still no #!

python/tox.ini
hpk42: sphinx released 2.0 onthe day i was sinking the many hours into build stuff, and it somehow breaks doc-generation.  1.8.5 works so i just fixated it.  time to worry about this for another day but we are not making heavy use of sphinx so no biggie i think.

ci_scripts/docker-coredeps/deps/build_rust.sh
hpk42: good catch.

r10s:
\o/

PR #645, opened 2019-04-08
New backup import export functions - less memory usage

csb0730:
- new backup file format which uses SQLITE _sqlar_ format
- messenger db is stored as blob in archive
- all blob files are stored with dir name in archive
- no need for vacuum of huge backup file at import
- memory consumption drops to approx 2x backup file size
- export is done only in new sqlar file format
- import is done with old and new file format (automatic detection)

comments welcome

Attacks core issue https://github.com/deltachat/deltachat-core/issues/261

### DIFF
--- a/ci_scripts/config.yml
+++ b/ci_scripts/config.yml
@@ -32,9 +32,6 @@ jobs:
   upload_docs_wheels:
     machine: True
     steps:
-      - run:
-          command: |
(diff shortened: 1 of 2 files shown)


csb0730:
not tested is: if this new backup file is to be unpacked by sqlite by standard implementation.
Background: additional config table is not foreseen in standard format !

PR #657, opened 2019-04-16
[wip] dc_get_locations() can return message-only-locations or full-track

r10s:
this pr aims to modify dc_get_locations() so that it takes a flag parameter. for now, i've only prototyped the flags as follows:

- if DC_GL_MESSAGES is set, dc_get_locations() returns all locations selected by chat/contact/timespan that contain messages
- if DC_GL_TRACKS is set, dc_get_locations() returns all known locations selected by chat/contact/timespan; this can be used to draw a track then. this includes the locations returned by DC_GL_MESSAGES

the last location selected by chat/contact/timespan is always added implicitly. if only this information is wanted, the flags parameter can be set to 0.

@cyBerta @nicodh makes sense? what do you think?


r10s:
will implement this as needed

PR #661, opened 2019-04-22
New backup import-export functions for lesser memory consumption

csb0730:
Characteristics:
- new backup file format which uses SQLITE sqlar format
- messenger db is stored as blob in archive
- all blob files are stored with dir name in archive
- no need for vacuum of huge backup file at import
- memory consumption drops to approx 2x backup file size
- export is done only in new sqlar file format
- import is done with old and new file format (automatic detection)

Comments welcome


csb0730:
Detected some problems in path handling at device at import!

Will update the PR next days.

r10s:
please note, that we're moving to rust currently, see https://github.com/deltachat/deltachat-core-rust . if things work out, the c-core will probably go away at all. just now, the c-core is frozen, so we do not add new features there unless absolutely necessary.

csb0730:
Not really a problem.
Changes are not so big and C code can serve as code template and moved to rust.

csb0730:
@dignifiedquire 
Maybe You can have a look meanwhile?

r10s:
@csb0730 i think, the current task is to get the rust-core stable at all. if this is done, we think about new features :)

r10s:
closing this pr for reasons metioned above

PR #665, opened 2019-04-24
workaround for PLAIN authentication not announced quirk

kalikaneko:


src/dc_smtp.c
```diff
@@ -208,6 +208,15 @@ int dc_smtp_connect(dc_smtp_t* smtp, const dc_loginparam_t* lp)
 				}
 				r = mailesmtp_auth_sasl(smtp->etpan, "PLAIN", hostname, NULL, NULL, NULL, lp->send_user, lp->send_pw, NULL);
 			}
+
+			/*
+			* There is a quirk with *some* Mailservers configuration in which PLAIN auth will not work if not 
+			* properly announced by the server, but it will happily proceed if we assume it is allowed.
+			*/
+			if (r == MAILSMTP_NO_ERROR) {
```
r10s: shouldn't this read `if (r != MAILSMTP_NO_ERROR)`?

kalikaneko:
After more careful debugging, I realized that building from current master actually fixes the problem. The current version released in pypi is broken, and the version I got from hpk's devpi is the one that had the quirk with the authentication - I'm guessing this must have fixed by later commits.

```pip install -i https://m.devpi.net/dc/master/  deltachat 
Looking in indexes: https://m.devpi.net/dc/master/
Collecting deltachat
  Using cached https://m.devpi.net/dc/master/%2Bf/0d3/2286bea913850/deltachat-0.9.0-cp35-cp35m-manylinux1_x86_64.whl```

for some reason 0.9.0 insists on trying the username workaround for smtp login and fails:

```
5.64 [imap-0x5a50acabb6a0] DC_EVENT_INFO(0,'SMTP-server mail.sindominio.net:587 STARTTLS-connected.')
5.65 [imap-0x5a50acabb6a0] DC_EVENT_INFO(0,'Trying SMTP-Login workaround "raksha"...')
5.65 [imap-0x5a50acabb6a0] DC_EVENT_ERROR_NETWORK(1,'SMTP-login failed for user raksha (Login failed)')
```

But anyway, all good now. Closing the PR.

src/dc_smtp.c
kalikaneko: you are right, this shouldn't be fixing the bug I was seeing

r10s:
@kalikaneko thanks very much for figuring this out :)

PR #671, opened 2019-04-28
prepare py 0.10.0 release

hpk42:


hpk42:
after "pip install -i https://m.devpi.net/dc/py10 deltachat" i get this error: 
```
>>> import deltachat
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/hpk/venv/py35/lib/python3.5/site-packages/deltachat/__init__.py", line 1, in <module>
    from deltachat import capi, const
ImportError: /home/hpk/venv/py35/lib/python3.5/site-packages/deltachat/capi.abi3.so: undefined symbol: dc_array_get_marker
```

@flub any idea what this could be?

flub:
Likely it is linking at runtime to an older version of the libdeltachat.so
then at build time.

PR #672, opened 2019-04-28
Document compiling/installing libetpan

ralphtheninja:
This is needed for the node bindings on linux in case libetpan is too old. Since we're referring to this repository from the node bindings I thought it was best to put the information in this readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ To install these on debian you can type:
     sudo apt install libetpan-dev libssl-dev libsqlite3-dev libsasl2-dev libbz2-dev zlib1g-dev
 ```
 
-If you run into connection problems with IMAP (error #43) it might be the case that the `libetpan` version on your system is too old. In this case we recommend compiling `libetpan`:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,26 +25,26 @@ do not require that you build the library manually:
 
 - [Windows building](https://github.com/deltachat/deltachat-core/issues/306)
 
-If you can help with advancing or adding to these efforts, be our guest. 
+If you can help with advancing or adding to these efforts, be our guest.
```
flub: I know you didn't write this.  But "be our guest" sounds really offensive like "do what you want, we we'll never care or help".  Something like "please do contribute" would be much friendlier.

README.md
```diff
@@ -94,12 +94,32 @@ which you may need to install (we have some fallbacks if you don't):
 - [libsasl](https://cyrusimap.org/sasl/)
 
 To install these on debian you can type:
+
 ```
     sudo apt install libetpan-dev libssl-dev libsqlite3-dev libsasl2-dev libbz2-dev zlib1g-dev
 ```
 
+If you run into connection problems with IMAP (error #43) it might be the case that the `libetpan` version on your system is too old. In this case we recommend compiling `libetpan`:
```
flub: I'd refer to https://github.com/dinhviethoa/libetpan/pull/330 here so that we know why we're recommending this and when we can remove it or how to clarify.  In particular it's only relevant if your libetpan links against gnutls.

Issue #126, opened 2018-03-06
remove Re: when using the subject as name for ad-hoc groups


r10s:

mbeko:
Should [prefixes in other languages](https://en.wikipedia.org/wiki/List_of_email_subject_abbreviations#Abbreviations_in_other_languages) be handled as well or is it better to stick to `Re:` and `Fwd:` first as long as we don't have tests (I think so based on another issue I saw)?

mbeko:
Good idea. This could be improved by finding the last occurence of `: ` and remove it with the preceding string. So whole chains of `Re: ` and `Fwd: ` will be removed and I assume `: ` is rarely part of the actual subject. Also, including the space avoids confusion with emotes.

mbeko:
GitHub seems to have removed the space, so just in case: There is a space after the colon in the search string 😉

testbird:
Better not to assume the colon might not be used in the subject, even if just in "new subject (was: this old subject).

maybe remove all repetitions using someting like "at the beginning, all permutations of known case insensitve strings, optionally separated with colons or spaces and with a space at the end? `"^([re|fwd]:* *) "` (not sure if it is valid regexp)

mbeko:
Then we're back at the question whether to check only for `Re:` and `Fwd:` or their translations, too. If we're to check for the specific words, then I indeed thought about something along the lines of that regex.

So, now the questions are:
* Check for the colons?
  * At the second or third position only?
  * At the last position in the string?
* Check for specific words?
  * Only `Re:` and `Fwd:`?
  * Also the translations?

r10s:
I would search for the _first_ colon, if it is at the 2nd or 3rd position, remove the colon and the text left of this colon and left-trim the rest. After that I would repeat this procedure until there are no more matches.

This way we would get rid of all Re: Fwd: AW: and so on stuff at the beginning of the subject.

Questions: are the 2 or 3 characters sufficient? has anyone seen a localized Re: or Fwd: prefix with more or less characters? If it is not sufficient, we may use a list. Btw. maybe we should move all this into a handy function (mrtools.c (which needs some cleanup some ime)) , seems as if it is needed at least at two positions.

Regarding RegEx: Currently there is no RegEx-library available and i think we should not add one for such a minor thing.

testbird:
> [repeat trimming up to] first colon, if it is at the 2nd or 3rd position

Good idea, works even in the "Re: new subject (was: old subject)".

[...]

PR #681, opened 2019-05-19
Subject prefix detection

mbeko:
This adds the functionality to remove known prefixes from e-mail subjects, providing a function to fix issue #126. I have [provided some information under this issue](https://github.com/deltachat/deltachat-core/issues/126#issuecomment-458292037) as well.

Please note that you still have to take care of calling the function where needed because wanted to keep the changes self-contained for easier review and integration. But I can add it if you give me a pointer to the proper place.

I have first tried to add unit tests for the functions in Python, so I made adaptations to make the Python tests run for me, added Pipenv for easier development and adapted the readme file accordingly. After discussion on the mailing list, though, we agreed that it's best to write the unit tests in C.

I have used Cmocka for these tests and added a script to run them. In a next step, someone knowledgeable in Meson should integrate them properly into the build.


r10s:
as already [mentioned in the belonging issue](https://github.com/deltachat/deltachat-core/issues/126#issuecomment-520312756), efforts go to core-rust in between.

@mbeko great job, anyway 👍  if we want to follow this direction, your code is a good base to start with. however, might also be we find a completely different solution for the subject of ad-hoc groups.

mbeko:
All right, just tell me when you have decided on this, so I can help to either port this code or implement another solution.

r10s:
@mbeko thanks! if you feel the need to code or fix something in rust in between - there a are [still several interesting tasks available](https://github.com/deltachat/deltachat-core-rust/issues) :)

PR #692, opened 2019-06-26
show asm independingly of show-emails settings

r10s:
closes https://github.com/deltachat/deltachat-android/issues/980


r10s:
just checked, this is already fixed in rust at https://github.com/deltachat/deltachat-core-rust/blob/master/src/dc_receive_imf.rs#L376

